PR #1, opened 2022-05-12
Update README + add support for .env files + add .gitignore + format …

Woetfin:
* Updated README documentation to give more setup details and explain how to run the commands as npm scripts
* Added support for .env files so we can define the environment variables per project in one time
* Added .gitignore to prevent secrets and installed dependencies tracking
* Formatted JS code with the 'Prettier' extension


Woetfin:
@TheSmartnik I've fixed the merge conflict that happened after the previous merge, would it be ok for you to add the dotenv dependency to support .env files or do you think this is not a good feature to add?

TheSmartnik:
@Woetfin thanks! Everything looks good but good you please remove formatting changes? We'll add eslinter with silverfin config later on

Woetfin:
@TheSmartnik I've removed the Prettier formatting from the files

TheSmartnik:
@Woetfin thanks!

PR #3, opened 2022-05-27
PST feedback

ihcene:
- Make HOST configurable to work with local repository
- Use axios defaults for base url and headers
- Make the cli complain about missing ENV variables
- Import and persist more fields

### DIFF
--- a/index.js
+++ b/index.js
@@ -5,7 +5,7 @@ const fs = require('fs');
 const RECONCILIATOIN_FIELDS_TO_SYNC = ["name_nl", "name_fr", "name_en", "auto_hide_formula", "text_configuration", "virtual_account_number", "reconciliation_type", "public", "allow_duplicate_reconciliations", "is_active", "tests"]
 
 createNewTemplateFolder = async function (handle) {
-  const relativePath = `./${handle}`
+  const relativePath = `./data/${handle}`
 
   fsUtils.createFolders(relativePath)
   testFile = { name: "test", content: "" }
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -2,9 +2,10 @@ const api = require('./sf_api')
 const fsUtils = require('./fs_utils')
 const fs = require('fs');
 
+const RECONCILIATOIN_FIELDS_TO_SYNC = ["name_nl", "name_fr", "name_en", "auto_hide_formula", "text_configuration", "virtual_account_number", "reconciliation_type", "public", "allow_duplicate_reconciliations", "is_active", "tests"]
```
TheSmartnik, line 5: `allow_duplicate_reconciliations`, `is_active` and `public` aren't exported in our API

index.js
```diff
@@ -2,9 +2,10 @@ const api = require('./sf_api')
...
+
 createNewTemplateFolder = async function (handle) {
-  const relativePath = `./${handle}`
-  const emptyCallback = () => {}
+  const relativePath = `./data/${handle}`
```
TheSmartnik, line 8: Why the path was changed to `data`?

TheSmartnik (commented):
And what about feature requests?

index.js
ihcene-silverfin: I created a branch on the monolith but forgot to open the MR
It's open now https://gitlab.silverfin.com/development/silverfin/-/merge_requests/15429

index.js
ihcene-silverfin: We need to group everything in a folder to add it to .gitignore. Other solution is to prefix folders.
Since users will play with the templates code and also the sf-toolkit code, I thought it's usefull to have the data code is ignored

TheSmartnik: @ihcene-silverfin one of the requests from PST was to subfolder everything under `reconciliations_texts`/`shared_parts`/`account_detail_templates`. I believe this requests conflicts with adding `data` superfolder

ihcene-silverfin: I will take that into account thank you

PR #4, opened 2022-06-12
Add ability to run tests

TheSmartnik:


index.js
```diff
@@ -99,4 +99,34 @@ persistReconciliationText = async function (handle) {
...
+  const testContent = fs.readFileSync(testPath, 'utf-8')
+
+  const testParams = { 'template': constructReconciliationText(handle), 'tests': testContent }
+
+  const testRunResponse = await api.createTestRun(testParams)
+  const testRunId = testRunResponse.data
+  let testRun = { 'status': 'started' }
+
+  while (testRun.status === 'started') {
+    await new Promise(resolve => setTimeout(resolve, 2000))
```
ihcene, line 115: Extract magic number to constant

index.js
```diff
@@ -99,4 +99,34 @@ persistReconciliationText = async function (handle) {
...
+
+    const response = await api.fetchTestRun(testRunId)
```
ihcene, line 117: we should maybe handle the catch case, either here or in the fetchTestRun

index.js
```diff
@@ -99,4 +99,34 @@ persistReconciliationText = async function (handle) {
...
+    testRun = response.data
+  }
+
+  if (testRun.status === 'completed') {
```
ihcene, line 121: To avoid the nested if you can invert this condition to have a guard condition

sf_api.js
```diff
@@ -31,4 +31,12 @@ updateReconciliationText = function(id, attributes) {
   return axios.post(`reconciliations/${id}`, attributes, axiosConfig())
 }
 
-module.exports = { fetchReconciliationTexts, updateReconciliationText, findReconciliationText }
+createTestRun = function(attributes) {
```
ihcene, line 34: Missing `const` and possible usage of arrow function with implicit return (since it's a one liner)

index.js
TheSmartnik: @ihcene could you please elaborate how can I invert it? I mean I just don't see it

sf_api.js
TheSmartnik: @ihcene I'd rather keep all functions consistent then some arrow functions and some not

ihcene: You are right

index.js
ihcene: Alright

PR #8, opened 2022-07-07
Liquid Tests: readme

AgustinSilverfin:
- Avoid overwriting existing YAML file.
- Rename YAML file.
- Create a readme file.

Would this be a good approach to create the readme? Having it stored in a different folder. Or should we avoid that ?
Also, I used a read and write approach instead of copying to avoid issues with file permissions.

### DIFF
--- a/fs_utils.js
+++ b/fs_utils.js
@@ -1,4 +1,5 @@
 const fs = require('fs');
+const path = require('path');
 
 createFolder = function (path) {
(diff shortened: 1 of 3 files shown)


TheSmartnik:
@AgustinSilverfin adding a README is a splendid idea! Thanks

dbobfin (approved):

PR #10, opened 2022-08-15
Authentication

AgustinSilverfin:
First Draft of an authentication module. To use it:
- You will need to set first your firm id as an env var: `export SF_FIRM_ID=...`
- You can run the authorisation process: `.node_modules/sf-toolkit/bin/cli.js authorize`
- You should be able now to run any of the existing commands
 
Some comments:
- Credentials are now saved in a json file, still to decide where to store them.
- My idea is to make the firm id a required argument in the CLI, that way we could potentially handle different firms where the user has access to.
- Moved sf_api to a separate directory.

### DIFF
--- a/README.md
+++ b/README.md
@@ -54,11 +54,23 @@ You can add this `scripts` section to the `packages.json`, where we define `silv
 
 ## Add your API credentials as environmental variables
 
-You could use a new `.env` local file, or add them to an existing file in your system (e.g. `~/.bash-profile` or `~/.bashrc`)
+You could use a new `.env` local file, or add them to an existing file in your system (e.g. `~/.bash-profile`, `~/.bashrc` or `~/.zshrc`)
 
 ```
 export SF_API_CLIENT_ID=...
 export SF_API_SECRET=...
+```
+
+If you are only using one firm, you could set it up as an environmental variable. That way, you won't need to pass the `--firm <firmid>` option every time you run a command.
+
+```
+export SF_FIRM_ID=...
+```
+
+In case you need to use a different host, you can also set it up as an environmental variable. By default it would be `https://live.getsilverfin.com`.
+
+```
+export SF_HOST=...
 ```
 
 # How to use it
@@ -84,7 +96,7 @@ yarn silverfin update-reconciliation --firm <firm-id> --handle <handle>
 ## Import all existing reconciliations
 
 ```
-yarn silverfin import-all-reconciliations --firm <firm-id> --handle <handle>
+yarn silverfin import-all-reconciliations --firm <firm-id>
 ```
 
 ## Import an existing shared part
@@ -102,7 +114,7 @@ yarn silverfin update-shared-part --firm <firm-id> --handle <handle>
 ## Import all existing shared parts
 
 ```
-yarn silverfin import-all-shared-parts --firm <firm-id> --handle <handle>
+yarn silverfin import-all-shared-parts --firm <firm-id>
 ```
 
 ## Run a Liquid Test
--- a/README.md
+++ b/README.md
@@ -6,12 +6,13 @@ A command-line tool for Silverfin template development.
 
 - Update your templates from the command-line while storing them in git repositories.
 - Run your Liquid Tests from the command-line.
+- Generate Liquid Tests from existing company files.
 
 # Setup & Basic Usage
 
 ## Prerequsites
 
-To use this CLI tool you need to first obtain a Silverfin API access_token
+To use this CLI tool you first need to register an API application within Silverfin. You will get a `client_id` and `secret`.
 
 ## Install package manager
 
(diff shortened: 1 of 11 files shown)


AgustinSilverfin:
- Updated cli.js and how we handle CLI arguments.
- Now firm id is a required argument for most commands, and I think it should also be more clear and flexible now how the different commands work (cli.js)
- Also, commands names where changed. You check the new [Readme](https://github.com/silverfin/sf-toolkit/blob/auth/README.md) or use the built-in `--help` in the CLI

README.md
```diff
@@ -27,82 +28,112 @@ Install `yarn` globally:
...
+
+```
+yarn init
+```
+
+Install `sf-toolkit` as a dependency of your project:
 
-You can either add the environment variables locally in the file so you only need to add them once, or define them directly in the terminal. 
+```
+yarn add https://github.com/silverfin/sf-toolkit.git
```
Oluwadamilareolusakin, line 42: We should eventually consider publishing an official npm package so they can add by doing something like `yarn add sf-toolkit` or `yarn add silverfin`

README.md
```diff
@@ -27,82 +28,112 @@ Install `yarn` globally:
...
-touch .env
+  "scripts": {
+    "silverfin": "node ./node_modules/sf_toolkit/bin/cli.js"
+  },
 ```
 
-Add the following variables in  .env:
+## Add your API credentials as environmental variables
+
+You could use a new `.env` local file, or add them to an existing file in your system (e.g. `~/.bash-profile` or `~/.bashrc`)
```
Oluwadamilareolusakin, line 57: May want to include `.zshrc` in the list

README.md
```diff
@@ -27,82 +28,112 @@ Install `yarn` globally:
...
+export SF_API_SECRET=...
 ```
 
-### Linux / Mac terminal
+If you are only using one firm, you could set it up as an environmental variable. That way, you won't need to pass the `--firm <firmid>` option every time you run a command.
 
 ```
-export SF_FIRM_ID=<firm_id>
-export SF_ACCESS_TOKEN=<access_token>
+export SF_FIRM_ID=...
```
Oluwadamilareolusakin, line 67: Okay for the first iteration, but you can also include a command in your package to set the firm/context they want to interact with, maybe `yarn silverfin -f firm_id_here` or `yarn silverfin -firm firm_id_here` which stores the current firm moving forward, it's more flexible than setting an env var or entering the firm id with every command

README.md
```diff
@@ -27,82 +28,112 @@ Install `yarn` globally:
...
+
+## Import all existing reconciliations
 
 ```
-"scripts": {
-    "new-recon": "./node_modules/sf_toolkit/bin/cli.js new",
-    "import-recon": "./node_modules/sf_toolkit/bin/cli.js import",
-    "update-recon": "./node_modules/sf_toolkit/bin/cli.js persistReconciliationText"
-  }
+yarn silverfin import-all-reconciliations --firm <firm-id> --handle <handle>
```
Oluwadamilareolusakin, line 99: What's the difference between this and the import-reconciliation command, they look like they take the same set of arguments. Does the latter import the first reconciliation it finds in the firm with that handle and this one imports all reconciliations in a firm with a given handle?

Oluwadamilareolusakin (commented):
The work done here is really great 👏  I'm curious about who the audience of this tool is and if what we should be building is SDKs instead of CLI tools, just curious about if this is meant to be used as a standalone tool and who is meant to be using it. If it's meant to be standalone, I'm curious as to why we are writing a yarn based tool as opposed to something more natural like `kubectl` or `gcloud` CLIs

README.md
AgustinSilverfin: Yes, indeed. `yarn` or `npm run` should be temporary solutions until we publish an npm package

README.md
AgustinSilverfin: Indeed, `import-all-reconciliations` is not developed yet, just assigned the command name as a placeholder. But it should import all the reconciliations existing in the firm. The intention is to be used when you are setting up your local repository, and you still have no reconciliations in it. Similar to `import-all-shared-parts`. But this shouldn't have the `--handle <handle>` argument, only `--firm <firmid>`

Woetfin:
Great work! 🙌  The authentication module in the current way it works can be merged as far as I'm concerned. 

Some things we can (potentially) still improve upon in a later stage:
* Automatically receive the authorization code & authorized firm id back in the CLI
* Use the authorization url that already contains the `firm_id` to skip the firm selection step `https://live.getsilverfin.com/f/{{firm_id}}/oauth/authorize?client_id={{client_id}}&redirect_uri={{callback_url}}&response_type=code&scope={{scope}}`
* Publish as an npm package and make the `silverfin` command global
* Test integration in the CI workflow

Woetfin (approved):

dbobfin (approved):

TheSmartnik (approved):
Great job 👍  Everything looks find IMHO

PR #11, opened 2022-08-19
import-all-reconciliations

AgustinSilverfin:
Refactored a bit index.js and fs_utils.js
Added command to import all reconciliations together (`import-all-reconciliations` )

Note: first we need to merge `auth` branch.


AgustinSilverfin:
Included in a new [PR](https://github.com/silverfin/sf-toolkit/pull/13)

PR #13, opened 2022-09-01
New functions for shared parts and refactoring

AgustinSilverfin:
- New function to import all reconciliations at once: `import-all-reconciliations`
- New functions to add & removed shared parts to reconciliations: `add-shared-part`, `remove-shared-part`,
- New function to get all the current shared parts used in one reconciliation: `shared-parts-used`.
- Function `createFiles` splitted into `createLiquidTestFiles` and `createTemplateFiles`.
- Get Test from Silverfin while importing a reconciliation and store it in a YAML file.
- Reconciliation id and handle added to `config.json`
- Code refactoring.
- Update function `persistReconciliationText`: try to get the reconciliation id from `config.json`. This will avoid making an extra API call.
- `uncaughtException` and `unhandledRejection`: ask the user to log the error details as an Issue in this repository.


Woetfin (changes requested):
`import-all-reconciliations` : 
* Receiving errors in the console when running this with firm ID 13827
```
TypeError: Cannot read properties of undefined (reading 'reduce')
    at storeImportedReconciliation (/home/woet/silverfin/sf-toolkit/index.js:38:51)
    at /home/woet/silverfin/sf-toolkit/index.js:92:5
    at Array.forEach (<anonymous>)
    at importExistingReconciliations (/home/woet/silverfin/sf-toolkit/index.js:91:24)
    at processTicksAndRejections (node:internal/process/task_queues:96:5)
```
* Should this also import marketplace reconciliations, don't immediately see a use-case for keeping these in the repo? (Error above might also be occurring because of these templates having hidden code?) 

`add-shared-part` / `remove-shared-part`:
* Uses -r / --reconciliation to specify the handle, but the other functions use the -h / --handle flag for the same information. Can we align this? 

`import-all-shared-parts`: ✔️ 

`shared-parts-used`: 
* ✔️ the function works
* But shouldn't we also be running `toolkit.refreshSharedPartsUsed(options.handle)` inside the actions `import-reconciliation` and `import-all-reconciliations` to have the shared parts dependencies automatically added when importing? 

Function createFiles splitted into createLiquidTestFiles and createTemplateFiles: ✔️ 
Get Test from Silverfin while importing a reconciliation and store it in a YAML file: ✔️ 
Reconciliation id and handle added to config.json: ✔️ 
Code refactoring: ✔️ 
Update function persistReconciliationText: try to get the reconciliation id from config.json. This will avoid making an extra API call: ✔️ 
uncaughtException and unhandledRejection: ask the user to log the error details as an Issue in this repository: ✔️

AgustinSilverfin:
> Should this also import marketplace reconciliations, don't immediately see a use-case for keeping these in the repo? (Error above might also be occurring because of these templates having hidden code?)

@Woetfin 
I don't think we can import/update reconciliations from Partners right now. Don't think the API allows that ? :thinking: 
I'll look into this issue, is probably related to the hidden code that you mention, and skip those?

>Uses -r / --reconciliation to specify the handle, but the other functions use the -h / --handle flag for the same information. Can we align this?

I though handle could be confused between reconciliation and shared part. But i'll change it to align with the rest.

>But shouldn't we also be running toolkit.refreshSharedPartsUsed(options.handle) inside the actions import-reconciliation and import-all-reconciliations to have the shared parts dependencies automatically added when importing?

I wasn't sure about this, since the refresh functionality look at `local` files, not new ones fetched from the API.
What I was considering was a new command to run all together (import all templates, import all shared parts, create all the links) -> Ideally this three should be used only once, and it could be at that moment (start point?)

AgustinSilverfin:
@Woetfin addressed the first two. 
Maybe for the last one we should discuss first which is going to be the right flow before going any further ?

Woetfin:
1. Reconciliations:
- Made the CLI a little less verbose
- Added try-catch blocks mostly now for error handling of missing path error
- Fixed bug & changed the description for templates with no code in main part 
- Added id as fallback folder name for templates with blank handle
2. Shared parts: 
- Fixed small typo for if (sharedParts.lenght == 0)

> I wasn't sure about this, since the refresh functionality look at local files, not new ones fetched from the API.
What I was considering was a new command to run all together (import all templates, import all shared parts, create all the links) -> Ideally this three should be used only once, and it could be at that moment (start point?)
--> Other changes look ok, I guess since they would only need to run these commands for all templates once, it's not worth putting more effort into combining this right now.

Woetfin (approved):
Can be merged if my changes are also ok for you @AgustinSilverfin

PR #16, opened 2022-10-12
Improve liquid test runs

Woetfin:


index.js
```diff
@@ -374,24 +376,40 @@ async function runTests(handle) {
```
AgustinSilverfin, line 376: the reconciliations liquid content that we provide is not including shared parts.

Function  constructReconciliationText(handle) only provides main and local parts, not shared parts. 

Is not clear in the [documentation](https://live.getsilverfin.com/api_docs#!/templates/postV4FRootFirmIdReconciliationsTest) as well how it should be handled

When I run a test from the CLI which has shared parts I get this error: 
`Liquid error: Liquid error (line 4): Unknown part shared/general_translations include "shared/general_translations"`
While running the same test in the Platform IDE, all tests pass.

index.js
```diff
@@ -374,24 +376,40 @@ async function runTests(handle) {
...
+      });
       const response = await SF.fetchTestRun(testRunId);
       testRun = response.data;
     }
 
+    spinner.stop();
+    readline.clearLine(process.stdout);
+    readline.cursorTo(process.stdout, 0);
+
     if (testRun.status !== "completed") {
```
AgustinSilverfin, line 399: Should we look at more precise codes ? 
Instead of `!== "completed"`, `=== "test_error"` ?
Do we have a list of all the possible status?

index.js
```diff
@@ -374,24 +376,40 @@ async function runTests(handle) {
...
       console.error(testRun.error_message);
       process.exit(1);
     }
 
     if (testRun.result.length !== 0) {
```
AgustinSilverfin, line 404: I wasn't able to test this. When it should happen ?

AgustinSilverfin (commented):
A general comment about the `fsErrorHandler()` function implemented earlier.

Now, since we are catching errors with this new function, I think we are no longer catching errors with the `uncaughtErrors()` function defined in `cli.js`. This function displays a message to report errors as issues in the repository.
I think we should remove the `else` condition  in `fsErrorHandler()` or move that message we display in `cli.js` here as well, into a new function that we can use it both in `fsErrorHandler()` and `uncaughtErrors()`

AgustinSilverfin (changes requested):
See comments left in the code

AgustinSilverfin:
The template information that we provide when we run the test is not including the "reconciliation_type" parameter

The `constructReconciliationText()` function is not including it. This is because we cannot push this parameter to Silverfin when we update the template.

index.js
```diff
@@ -374,24 +376,40 @@ async function runTests(handle) {
...
       console.error("Tests Failed");
-      console.error(testRun.result);
+      console.dir(testRun.result, { maxArrayLength: null });
       process.exit(1);
+    } else {
+      if (testRun.status == "completed") {
+        console.log("ALL TESTS PASSED");
+        process.exit(1);
```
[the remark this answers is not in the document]
Woetfin: Updated

index.js
```diff
@@ -374,24 +376,40 @@ async function runTests(handle) {
 
     const testRunResponse = await SF.createTestRun(testParams);
     const testRunId = testRunResponse.data;
+    console.log(testRunId);
```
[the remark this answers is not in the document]
Woetfin: Removed; was still there for dev purposes

index.js
```diff
@@ -1,6 +1,8 @@
 const SF = require("./api/sf_api");
 const fsUtils = require("./fs_utils");
 const fs = require("fs");
+const readline = require("readline");
+const Spinner = require("cli-spinner").Spinner;
```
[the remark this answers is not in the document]
Woetfin: I've removed any package for the spinner and created a custom one; it's pretty barebones now but at least it's something

index.js
```diff
@@ -374,24 +376,40 @@ async function runTests(handle) {
...
       const response = await SF.fetchTestRun(testRunId);
       testRun = response.data;
     }
 
+    spinner.stop();
+    readline.clearLine(process.stdout);
+    readline.cursorTo(process.stdout, 0);
+
     if (testRun.status !== "completed") {
       console.error(testRun.error_message);
```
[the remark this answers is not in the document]
Woetfin: I've done a major overhaul of all the displaying of the test results

index.js
Woetfin: Would need to ask ENG for a list cause not sure if there's a specific error code

index.js
Woetfin: Was still a snippet added by ENG before; but I think this should trigger if the test run fails on technical level (internal server error or a timeout error)

AgustinSilverfin (approved):

PR #21, opened 2022-12-08
New CLI command: import reconciliation by id

AgustinSilverfin:
Add the option to import reconciliations with the ID instead of the handle
`silverfin import-reconciliation --id <id>`

### DIFF
--- a/api/sf_api.js
+++ b/api/sf_api.js
@@ -232,6 +232,25 @@ async function findReconciliationText(firmId, handle, page = 1) {
   }
 }
 
+async function findReconciliationTextById(firmId, reconciliationId, page = 1) {
+  const response = await fetchReconciliationTexts(firmId, page);
+  const reconciliations = response.data;
+  // No data
+  if (reconciliations.length == 0) {
+    console.log(`Reconciliation ${reconciliationId} not found`);
+    return;
+  }
+  let reconciliationText = reconciliations.filter(
+    (element) => element["id"] === Number(reconciliationId)
+  )[0];
+  // Only return reconciliations were liquid code is not hidden
+  if (reconciliationText && reconciliationText.hasOwnProperty("text")) {
+    return reconciliationText;
+  } else {
+    return findReconciliationText(firmId, reconciliationId, page + 1);
+  }
+}
+
 async function updateReconciliationText(
   firmId,
   reconciliationId,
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -110,13 +110,24 @@ function storeImportedReconciliation(reconciliationText) {
 }
 
 async function importExistingReconciliationByHandle(firmId, handle) {
-  reconciliationText = await SF.findReconciliationText(firmId, handle);
+  let reconciliationText = await SF.findReconciliationText(firmId, handle);
```
Woetfin, line 113: Clearer to make this a `const` since the value is not being overwritten?

index.js
```diff
@@ -110,13 +110,24 @@ function storeImportedReconciliation(reconciliationText) {
...
   if (!reconciliationText) {
     throw `${handle} wasn't found`;
   }
   storeImportedReconciliation(reconciliationText);
 }
 
+async function importExistingReconciliationById(firmId, reconciliationId) {
+  let reconciliationText = await SF.findReconciliationTextById(
```
Woetfin, line 121: Clearer to make this a `const` since the value is not being overwritten?

Woetfin (changes requested):
See comments in code

api/sf_api.js
```diff
@@ -232,6 +232,29 @@ async function findReconciliationText(firmId, handle, page = 1) {
...
+  // No data
+  if (reconciliations.length == 0) {
+    console.log(`Reconciliation ${reconciliationId} not found`);
+    return;
+  }
+  let reconciliationTexts = reconciliations.filter(
+    (element) => element["id"] === Number(reconciliationId)
+  );
+  if (reconciliationTexts.length != 0) {
+    for (let reconciliationText of reconciliationTexts) {
```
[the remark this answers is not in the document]
AgustinSilverfin: Valid point, we still need to check if it has text (hidden templates). I'll adapt it

api/sf_api.js
```diff
@@ -232,6 +232,25 @@ async function findReconciliationText(firmId, handle, page = 1) {
...
 
+async function findReconciliationTextById(firmId, reconciliationId, page = 1) {
+  const response = await fetchReconciliationTexts(firmId, page);
+  const reconciliations = response.data;
+  // No data
+  if (reconciliations.length == 0) {
+    console.log(`Reconciliation ${reconciliationId} not found`);
+    return;
+  }
+  let reconciliationText = reconciliations.filter(
```
Woetfin, line 243: Should be a `const` too here

Woetfin (approved):
Small change from let to const, can still be made, but can be merged imo

PR #22, opened 2022-12-12
Fix: exit liquid test run process when an error occurred

Woetfin:
* Exit running liquid test process when an error occurred
* Clean spinner
* Disable keyboard when spinner is running (still to be added)


AgustinSilverfin:
We cannot add process.exit, the function should return something and continue. 
That's why I created the class with the start/stop methods.
It is picked by the extension for example. If not, we should create separate functions

AgustinSilverfin:
Can we close this PR?

PR #29, opened 2023-01-19
Check CLI latest version

AgustinSilverfin:
## Description

Implement some functionality to check the CLI version. Look for the latest version (package.json in the github repo) and compare it with your current version.
If there is an update available, display a message suggesting to update.
Also add a script  command with a more user-friendly way of udpating: 
`silverfin update` instead of `npm install -g https://github.com/silverfin/sf-toolkit.git`

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -6,7 +6,9 @@ const stats = require("../stats_utils");
 const { Command } = require("commander");
 const prompt = require("prompt-sync")({ sigint: true });
 const pkg = require("../package.json");
+const cliUpdates = require("../resources/cliUpdates");
 const program = new Command();
+const { exec } = require("child_process");
 
 // Load default firm id from Config Object or ENV
 let firmIdDefault = undefined;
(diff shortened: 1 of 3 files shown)


bin/cli.js
```diff
@@ -355,4 +359,34 @@ program
...
 
-program.parse();
+// Update the CLI
+if (pkg.repository && pkg.repository.url) {
+  program
+    .command("update")
+    .description("Update the CLI to the latest version")
+    .action(() => {
+      console.log(`Updating npm package from GitHub repository...`);
+      exec(
```
Woetfin, line 369: Can be a long running command; better to also add the spinner here to give the user feedback the process is running?

Woetfin (approved):
Would add the spinner to provide feedback to the user while silverfin update is running; but the general functionality is working so can be merged if you don't deem it necessary. 
Also could use some colors in the message with the chalk package that's already being used so the message is harder to ignore: 
W31V0

bin/cli.js
AgustinSilverfin: I tried to add the spinner, but it was conflicting with the user input required to be typed into the terminal (password). I'll skip it for now

PR #30, opened 2023-02-06
Liquid test html

AgustinSilverfin:
## Description

- Adapt the Liquid Test run flow to support the new API response structure
- Add support for Liquid Test HTML response

## Type of change

- [ ] Bug fix
- [X] New feature
- [X] Breaking change

## Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -281,9 +281,15 @@ program
     "-h, --handle <handle>",
     "Specify the reconciliation to be used (mandatory)"
   )
+  .option(
+    "-t, --test <test-name>",
+    "Specify the name of the test to be run (optional)",
+    ""
+  )
   .action((options) => {
     checkDefaultFirm(options.firm);
-    toolkit.runTests(options.firm, options.handle);
+
+    toolkit.runTests(options.firm, options.handle, options.test);
   });
 
 // Create Liquid Test
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -286,10 +286,19 @@ program
     "Specify the name of the test to be run (optional)",
     ""
   )
+  .option(
+    "--html",
+    "Get a html file of the template generated with the Liquid Test information (optional)",
+    false
+  )
   .action((options) => {
     checkDefaultFirm(options.firm);
-
-    toolkit.runTests(options.firm, options.handle, options.test);
+    toolkit.runTestsWithOutput(
+      options.firm,
+      options.handle,
+      options.test,
+      options.html
+    );
   });
 
 // Create Liquid Test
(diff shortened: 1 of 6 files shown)


index.js
```diff
@@ -395,229 +402,327 @@ async function removeSharedPartFromReconciliation(
...
+  } catch (err) {}
+}
+
+async function handleHTMLfiles(testName = "", testRun) {
+  deleteExistingHTMLs();
+  if (testName) {
+    // Only one test
+    await getHTML(testRun.tests[testName].html, testName, true);
+  } else {
+    // All tests
```
Woetfin, line 688: Should we make it not open all tests by default in the browser if there's more than one defined? (Add a new flag like `--open-all` / `-o`)

Woetfin (approved):
* I added a new commit that removed any testing `console.log` statement still in the code
* I also added support to open the HTML when working with WSL (probably an edge case and didn't seem super straighforward to support, but I'm working with WSL most of the time so is useful to support :D), the open package didn't seem to support WSL properly. 
* Other comments are open questions, but I don't think they should block release so can go live imo if you're ok with my change. 
* Added some feedback for ENG for HTML return here still: https://gitlab.silverfin.com/development/silverfin/-/merge_requests/16379#note_868006, so once they answer that I think we can go live.

bin/cli.js
```diff
@@ -281,9 +281,24 @@ program
     "-h, --handle <handle>",
     "Specify the reconciliation to be used (mandatory)"
   )
+  .option(
+    "-t, --test <test-name>",
+    "Specify the name of the test to be run (optional)",
+    ""
+  )
+  .option(
+    "-e, --html",
```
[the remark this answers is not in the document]
AgustinSilverfin: export, but maybe it's better to remove it and keep only `--html` (`-h` was taking for `--handle`)

index.js
AgustinSilverfin: I would say if you pass the `--html` flag, you would like to open them in the browser. If not, if you are not storing them anywhere so there is no point of using `--html`?

AgustinSilverfin:
>I added a new commit that removed any testing console.log statement still in the code
I don't see any `console.log` being removed in your commit :)

bin/cli.js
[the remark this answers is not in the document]
Woetfin: yeah, maybe just having --html is short enough and clear in all cases

index.js
Woetfin: Idd, we can leave it as is

Woetfin:
> > I added a new commit that removed any testing console.log statement still in the code
> 
> I don't see any `console.log` being removed in your commit :)

I guess they were still in there from myself on my local branch when I took my first look at the PR a while ago 😅

Woetfin (approved):
Tested on WSL & Mac and all working as expected

PR #39, opened 2023-05-31
Uk 4216 dev liquid repo align create reconciliation with import reconciliation

Toby-Masters-SF:
## Description

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/fs_utils.js
+++ b/fs_utils.js
@@ -117,10 +117,15 @@ function createConfigIfMissing(relativePath, templateType = undefined) {
     const config = { id: {} };
     if (templateType == "reconciliation_text") {
       config.text = "main.liquid";
-      config.text_parts = {};
       config.auto_hide_formula = "";
       config.text_configuration = null;
       config.virtual_account_number = "";
+      config.reconciliation_type = "only_reconciled_with_data";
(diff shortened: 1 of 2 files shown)


fs_utils.js
```diff
@@ -117,10 +117,15 @@ function createConfigIfMissing(relativePath, templateType = undefined) {
...
       config.text = "main.liquid";
-      config.text_parts = {};
       config.auto_hide_formula = "";
       config.text_configuration = null;
       config.virtual_account_number = "";
+      config.reconciliation_type = "only_reconciled_with_data";
+      config.public = false;
+      config.allow_duplicate_reconciliation = false;
+      config.is_active = true;
+      config.externally_managed == false;
```
AgustinSilverfin, line 127: I think this should be set to `true` by default.
Now is not working since there is an issue in the API endpoint, but once it's fixed we will be setting this to true to every template ideally

Woetfin: @Toby-Masters-SF should also only contain one `=` here to assign the variable variable

Woetfin (changes requested):

Issue #37, opened 2023-05-17
CLI: reducing printed information when using import-shared-part command


MelSFNL:
**Is your feature request related to a problem? Please describe.**
When importing all shared parts from a firm, a lot of API call responses etc. are printed. This is not informative and floods the terminal. 

**Describe the solution you'd like**
Either:
(1) Add a --verbose option which reduces the printed information when using the import-shared-part command
or:
(2) Make the --verbose option standard and include another option to print all info (i.e. the other way around)

**Describe alternatives you've considered**
None

**Additional context**
Small and insignificant feature request, but needed to clear it from my to do list

AgustinSilverfin:
The reason why we started like this (verbose by default) was to make explicit every time that we were making an API request. So that way we could be conscious while using the CLI and not abuse of those requests.
But I agree with you,  it's one of the things that I would like to change as well, whenever we have the time to do it.

PR #50, opened 2023-08-04
New API methods for export files

AgustinSilverfin:
## Description

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist


Woetfin:
Shouldn't we split this up in separate PRs?

Maybe we can then release each part in this order and already confirm that those parts are running well: 
1. Refactor templates into Classes. Align method names.
2. Add support for account details.
3. Add support for export files.
4. Add --message flag for update and create commands

AgustinSilverfin:
I can split it like that as long as the refactoring is the first one :slightly_smiling_face: 
So, let's focus on getting this done first #55 
And then I'll split this one into separate PRs

Woetfin:
@AgustinSilverfin Sounds good! 

I think this request might also be a good one to pick up during the refactoring? 
https://github.com/silverfin/silverfin-cli/issues/37

index.js
```diff
@@ -111,6 +113,111 @@ async function newAllReconciliations(firmId) {
...
+  }
+}
+
+async function fetchAccountDetailByHandle(firmId, name) {}
+async function fetchAccountDetailById(firmId, id) {}
+async function fetchAllAccountDetails(firmId) {}
+async function publishAccountDetailByName(firmId, name) {}
+async function publishAllAccountDetails(firmId) {}
+async function newAccountDetail(firmId, name) {}
+async function newAllAccountDetails(firmId) {}
```
Woetfin, line 219: Should we leave these as part of this PR?

index.js
```diff
@@ -211,114 +318,99 @@ async function newAllSharedParts(firmId) {
   }
 }
 
-/** This function adds a shared part to a reconciliation text. It will make a POST request to the API. If the ID of one of the templates it missing, it will try to fetch it first by making a GET request. In case of success, it will store the details in the corresponding config files.
+/** This function adds a shared part to a template. It will make a POST request to the API. If the ID of one of the templates it missing, it will try to fetch it first by making a GET request. In case of success, it will store the details in the corresponding config files.
```
Woetfin, line 321:
```suggestion
/** This function adds a shared part to a template. It will make a POST request to the API. If the ID of one of the templates is missing, it will try to fetch it first by making a GET request. In case of success, it will store the details in the corresponding config files.
```

small typo

index.js
```diff
@@ -327,65 +419,86 @@ async function addSharedPartToReconciliation(
...
 /**
- * This function loops through all shared parts (config files) and tries to add the shared part to each reconciliation listed in 'used_in'. It will make a POST request to the API. If the ID of one of the templates it missing, it will try to fetch it first by making a GET request. In case of success, it will store the details in the corresponding config files.
+ * This function loops through all shared parts (config files) and tries to add the shared part to each template listed in 'used_in'. It will make a POST request to the API. If the ID of one of the templates it missing, it will try to fetch it first by making a GET request. In case of success, it will store the details in the corresponding config files.
  * @param {Number} firmId
  */
-async function addAllSharedPartsToAllReconciliation(firmId) {
+async function addAllSharedParts(firmId) {
   const sharedPartsArray = fsUtils.getAllTemplatesOfAType("sharedPart");
-  for (let sharedPartName of sharedPartsArray) {
+  for await (let sharedPartName of sharedPartsArray) {
```
Woetfin, line 427: I think the for .. of loop is already synchronous, so the await keyword isn't necessary

Woetfin (changes requested):
Liking the organisation of the template types in separate classes 👍 making we can add those to a separate "templateTypes" directory as well inside the "lib" directory? 

Tested all commands:
1. import-export-file
	1. Change to only support import with id so it's aligned with the account templates? 
	2. function checkValidName ==> for export files / account templates, spaces and other characters should be allowed since the name of the folder will be a name that could contain this and be valid. 
	3. the import also creates a "tests" folder, but since there are no liquid tests for export files, it can be skipped. 
2. update-export-file
	1. Looks good
3. development-mode
	1. Update from export file doesn't seem to trigger update from dev mode 
4. create-export-file
	1. the create also creates a "tests" folder, but since there are no liquid tests for export files, it can be skipped. 
5. add-shared-part
	1. Looks good
6. remove-shared-part
        1. Looks good
7. get-export-file-id
	1. Looks good

index.js
AgustinSilverfin: Removed

index.js
AgustinSilverfin: I think we actually needed. Not sure if both awaits :thinking:

AgustinSilverfin:
@Woetfin I think I have addressed all your comments. Could you test again?
As we discussed, I kept the import with name as consistency (since you can update and create with name)

index.js
Woetfin: I did a test and I believe it's finishing the first loop and doing it in sequential order, so I don't think the await is necessary here; checking this documentation https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for-await...of ; I believe it still works cause we're using it to loop over 'sync iterables', but we can do that as well with the normal one: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/for...of

AgustinSilverfin: Removed

Woetfin (approved):
Tested commands for the export files import / update / add shared part / remove shared part / development mode and all working well, so I think it's good to go 🔥

PR #54, opened 2023-08-21
add User-Agent header to requests

Woetfin:
## Description

Added User-Agent header requests to engineering for easier tracking which requests originate from the CLI. 

We should now be sending these headers on a request:
e.g. POST /api/v4/f/13827/reconciliations/67081995 HTTP/1.1

```
Accept: application/json, text/plain, */*
Authorization: Bearer <access_token>
Content-Type: application/json
User-Agent: silverfin-cli/1.14.2
Content-Length: 808
Host: live.getsilverfin.com
Connection: close
```

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/api/sfApi.js
+++ b/lib/api/sfApi.js
@@ -1,6 +1,7 @@
 const axios = require("axios");
 const open = require("open");
 const prompt = require("prompt-sync")({ sigint: true });
+const pkg = require("../../package.json");
 const { firmCredentials } = require("./firmCredentials");
 
 const baseURL = process.env.SF_HOST || "https://live.getsilverfin.com";
@@ -117,6 +118,7 @@ function setAxiosDefaults(firmId) {
   if (firmTokens) {
     axios.defaults.baseURL = `${baseURL}/api/v4/f/${firmId}`;
     axios.defaults.headers.common[
+      "User-Agent" = `Silverfin CLI/${pkg.version}`,
       "Authorization"
     ] = `Bearer ${firmTokens.accessToken}`;
   } else {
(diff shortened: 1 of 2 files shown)


lib/api/sfApi.js
```diff
@@ -117,6 +118,7 @@ function setAxiosDefaults(firmId) {
   if (firmTokens) {
     axios.defaults.baseURL = `${baseURL}/api/v4/f/${firmId}`;
     axios.defaults.headers.common[
+      "User-Agent" = `Silverfin CLI/${pkg.version}`,
       "Authorization"
     ] = `Bearer ${firmTokens.accessToken}`;
```
AgustinSilverfin, line 123: The previous code doesn't work, it will break the CLI. The User-Agent should be outside `common`

```suggestion
    axios.defaults.baseURL = `${baseURL}/api/v4/f/${firmId}`;
    axios.defaults.headers["User-Agent"] = `Silverfin CLI/${pkg.version}`;
    axios.defaults.headers.common[
      "Authorization"
    ] = `Bearer ${firmTokens.accessToken}`;
```

AgustinSilverfin (changes requested):
This will be a nice addition, we can use it to measure usage as well :+1:

lib/api/sfApi.js
Woetfin: Moved the line outside common 
Also used `silverfin-cli` naming instead so it also matches our project name

AgustinSilverfin (approved):
:+1:

PR #55, opened 2023-08-24
Support HTML render for input and/or preview mode from new /reconciliations/render endpoint

Woetfin:
## Description

Update the CLI `run-test` command to take new options;
`--html-input`: render only the input view
`--html-preview`: render only the export view

Both these options can also be combined to render both views at the same time. 

`--skip-results`: if we're only interested in rendering the preview

We can do this now since it's on a separate endpoint. 

So we're no longer required to always render results and possibly throw errors if that's not what the developer is focusing on 

If they forget to run the test results output, the GitHub Action will fail either way so there shouldn't be risk that they overlook the liquid testing results.

## Type of change

- [ ] Bug fix
- [x] New feature
- [x] Breaking change

## Checklist


lib/liquidTestRunner.js
```diff
@@ -326,13 +330,17 @@ async function runTestsWithOutput(
...
   handle,
   testName = "",
-  html_render = false
+  htmlInput = false,
+  htmlPreview = false
 ) {
   try {
-    const testRun = await runTests(firmId, handle, testName, html_render);
+    // console.log("htmlInput", htmlInput);
+    // console.log("htmlPreview", htmlPreview);
```
AgustinSilverfin, line 338: Let's remove this debug variables

lib/utils/runTestUtils.js
```diff
@@ -0,0 +1,15 @@
+function checkMode(htmlInput, htmlPreview) {
```
AgustinSilverfin, line 1: Maybe let's rename the function to something mentioning html: `checkHtmlMode`

lib/cli/devMode.js
```diff
@@ -37,12 +37,7 @@ async function watchLiquidTest(firmId, handle, testName, html_render) {
     lastUpdates[filename] = fileStats.mtimeMs;
 
     // Run test
-    await liquidTestRunner.runTestsWithOutput(
-      firmId,
-      handle,
-      testName,
-      html_render
-    );
+    await liquidTestRunner.runTestsWithOutput(firmId, handle, testName, mode);
```
AgustinSilverfin, line 40: `runTestsWithOutput` accepts `htmlInput` and `htmlOutput` as parameters, not `mode`.
Actually I think like more the  idea of passing `mode` and doing the mode check outside the function ?
Also, we could rename `mode` to `htmlMode` in the parameters

bin/cli.js
```diff
@@ -281,8 +281,18 @@ program
...
+    "Get a static html of the input-view of the template generated with the Liquid Test data (optional)",
+    false
+  )
+  .option(
+    "--html-preview",
+    "Get a static html of the export-view of the template generated with the Liquid Test data (optional)",
+    false
+  )
+  .option(
+    "--skip-results",
```
AgustinSilverfin, line 294: You are not using `skip-results`

AgustinSilverfin: `--skip-results` will be quite similar to `--status` (which only displays PASSED/FAILED status back). 
Do we need this slight variant?

AgustinSilverfin (changes requested):

AgustinSilverfin:
@Woetfin I've just realized it is a draft PR :sweat_smile: 
Since I was tagged to do a review and github showing a banner at the top asking for it, I haven't noticed it. But OK, comments are still valid

bin/cli.js
Woetfin: I've renamed it to `--html-only`; I don't think it will be similar though, cause status gives ONLY final liquid test output, while this will give ONLY preview output.

lib/cli/devMode.js
Woetfin: I think this is something I hadn't gotten to yet cause it was in draft

lib/liquidTestRunner.js
Woetfin: console logs will be removed when I've finalized the dev

lib/utils/runTestUtils.js
Woetfin: Will rename to checkRenderMode (cause this might be rendering PDFs in the future too)

lib/liquidTestRunner.js
```diff
@@ -308,9 +312,9 @@ async function handleHTMLfiles(testName = "", testRun) {
 }
 
 // Used by VSCode Extension
-async function runTests(firmId, handle, testName = "", html_render = false) {
+async function runTests(firmId, handle, testName = "", mode) {
```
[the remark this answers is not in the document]
Woetfin: mode will always return something from the checkRenderMode function, so it's kind of the same as adding a default here?

lib/cli/devMode.js
AgustinSilverfin: I still think this is not correct: `mode` refers to `previewOnly` parameter? I would at least rename it to make it more clear.
So when we use `--html` in `development-mode` we want only the HTML (from input?) and not the output in the terminal ?
I'm getting this error when I add the `--html` flag

Screenshot from 2023-09-12 12-30-53

Also, if we use `--preview-only` we must use at least one of `--html-input` or `--html-preview`. If non of them are added which preview will we take?

I get exactly the same error as before (and I think that's the issue in `development-mode`) while running:
`silverfin run-test -h handle --preview-only`

lib/api/sfApi.js
```diff
@@ -485,6 +485,29 @@ async function createTestRun(firmId, attributes, refreshToken = true) {
...
+    const callbackParameters = {
+      firmId,
+      attributes,
+      refreshToken: false,
+    };
+    const response = await apiUtils.responseErrorHandler(
+      firmId,
+      error,
+      refreshToken,
+      createTestRun,
```
AgustinSilverfin, line 503: this should be `createPreviewRun`

AgustinSilverfin (changes requested):

lib/liquidTestRunner.js
```diff
@@ -297,30 +307,78 @@ async function deleteExistingHTMLs() {
...
     testNames.forEach(async (testName) => {
-      await getHTML(testRun.tests[testName].html, testName, true);
+      getHTMLrenders(mode, testName, testRun, true);
     });
   }
 }
 
 // Used by VSCode Extension
-async function runTests(firmId, handle, testName = "", html_render = false) {
+async function runTests(firmId, handle, testName = "", previewOnly, mode) {
```
AgustinSilverfin, line 361: I think we should approach this differently, since there is no point of having `previewOnly` as `true` when `mode` is `none` ?
I think we should do those check outside the function and pass the correct option (so, `mode` only actually would have all the information) -> and rename it to `htmlMode` or somethimg more clear

AgustinSilverfin (changes requested):

lib/liquidTestRunner.js
Woetfin: I'm not sure I follow, `previewOnly` is also a flag they can add to the command (`--preview-only`) and doesn't have any direct link with `mode`.

They can idd still add no HTML render option and `--preview-only` flag, so in that case, the flag would be redundant since mode is already `none`, but I didn't really see that as an issue? 

This is the fastest option I could think of on how to cover the case where want to render HTML without any liquid testing results, but also still give the user the option to combine both. 

Cause now I wanted to cover all these use-cases:
* Get liquid test results only: use `run-test` command without any flag
* Get liquid test results + HTML render: use `run-test` with `--html-input` and/or `--html-preview` flag
* Get HTML render only: use `run-test` with `--html-input` and/or `--html-preview` flag and the `--preview-only` flag. 

Unless you see a better way of tackling these use-cases?

AgustinSilverfin: Related to this comment: https://github.com/silverfin/silverfin-cli/pull/55#discussion_r1322846486. 
We can have a quick chat if it's not clear.
tl;dr: `--preview-only` for `run-tests` or `--html` for `development-mode` raise an error for me
i still think it's a bit confusing too many options an parameters and we can do it before calling the different run tests functions, but not really required.

lib/api/sfApi.js
Woetfin: updated

lib/cli/devMode.js
Woetfin: This should be working now, `development-mode` will render only input-mode of the template. I think the auto-rendering of the input/preview HTMLs will be cleaner to integrate in the extension for now or already have the option by using `run-test` manually. 

Cause when using the CLI `development-mode` for rendering the different HTMLs on the liquid tests, it really spams you with tabs so I'd also want to fix the opening of the page in a new tab every time when integrating this completely in this command (so maybe better to take that into a separate PR)

lib/liquidTestRunner.js
Woetfin: The only way I could think about on how to make it more clear is if we also do a bigger change in the CLI commands and split run-tests / render into separate processes.

AgustinSilverfin:
@Woetfin I added a check so `preview-only` is not used alone. I know it is not supposed to used alone, but it was still raising an error if you don't use it properly.
Also, renamed a bit all the new mode parameters (to renderMode or htmlMode depending cases)
Feel free to do a final test and merge, looks good to me :+1:

AgustinSilverfin (approved):

PR #56, opened 2023-08-29
Refactor code

AgustinSilverfin:
## Description

Rename several methods to align across template types 
Create separate classes for each template type
Preparation before adding support for export files and account details

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

- [] README updated (if needed)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -9,3 +9,4 @@ shared_parts
 account_templates
 permanent_texts
 export_files
+account_details
--- a/lib/api/sfApi.js
+++ b/lib/api/sfApi.js
@@ -1,9 +1,8 @@
 const axios = require("axios");
 const open = require("open");
 const prompt = require("prompt-sync")({ sigint: true });
-const { firmCredentials } = require("./firmCredentials");
+const apiUtils = require("../utils/apiUtils");
 
-const baseURL = process.env.SF_HOST || "https://live.getsilverfin.com";
 const missingVariables = ["SF_API_CLIENT_ID", "SF_API_SECRET"].filter(
   (key) => !process.env[key]
 );
@@ -35,7 +34,7 @@ async function authorizeApp(firmId = undefined) {
     } else {
       firmIdPrompt = prompt("Enter the firm ID: ");
     }
-    const url = `${baseURL}/f/${firmIdPrompt}/oauth/authorize?client_id=${process.env.SF_API_CLIENT_ID}&redirect_uri=${redirectUri}&response_type=code&scope=${scope}`;
+    const url = `${apiUtils.BASE_URL}/f/${firmIdPrompt}/oauth/authorize?client_id=${process.env.SF_API_CLIENT_ID}&redirect_uri=${redirectUri}&response_type=code&scope=${scope}`;
 
     await open(url);
     console.log(`You need to authorize your APP in the browser`);
@@ -44,7 +43,7 @@ async function authorizeApp(firmId = undefined) {
       echo: "*",
     });
     // Get tokens
-    const tokens = await getAccessToken(firmIdPrompt, authCodePrompt);
+    const tokens = await apiUtils.getAccessToken(firmIdPrompt, authCodePrompt);
     if (tokens) {
       console.log("Authentication successful");
     }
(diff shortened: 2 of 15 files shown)


lib/utils/apiUtils.js
```diff
@@ -0,0 +1,154 @@
...
+
+const BASE_URL = process.env.SF_HOST || "https://live.getsilverfin.com";
+
+// Get Tokens for the first time
+async function getAccessToken(firmId, authCode) {
+  try {
+    const redirectUri = "urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob";
+    const grantType = "authorization_code";
+    let requestDetails = {
+      method: "post",
```
Woetfin, line 13: Add HTTP verbs in all caps according to convention?

lib/api/sfApi.js
```diff
@@ -214,34 +78,56 @@ async function createReconciliationText(
   }
 }
 
-async function fetchReconciliationTexts(firmId, page = 1, refreshToken = true) {
-  setAxiosDefaults(firmId);
+async function readReconciliationTexts(firmId, page = 1, refreshToken = true) {
+  apiUtils.setAxiosDefaults(firmId);
```
Woetfin, line 82: I think we can actually remove the need to set the default settings each time with a function and create an axios instance in the utils that already has the default settings and send around that instance instead. 

The setAxiosDefaults function is pretty ingrained already though in our code so would be need to test everything if we change it, but I think using an instance would be a perfect use-case for what we're doing here.

lib/api/sfApi.js
```diff
@@ -214,34 +78,56 @@ async function createReconciliationText(
...
-  const reconciliations = response.data;
+async function readReconciliationTextById(firmId, id, refreshToken = true) {
+  apiUtils.setAxiosDefaults(firmId);
+  try {
+    const response = await axios.get(`reconciliation/${id}`);
+    apiUtils.responseSuccessHandler(response);
+    return response;
+  } catch (error) {
+    const callbackParameters = {
+      firmId: firmId,
```
Woetfin, line 114: No need to define property a property value if the name of the value & property match, so only `firmId` would work here too.

lib/cli/devMode.js
```diff
@@ -47,6 +47,7 @@ async function watchLiquidTest(firmId, handle, testName, html_render) {
 
   // Watch liquid files
   const liquidFiles = fsUtils.listExistingRelatedLiquidFiles(firmId, handle);
+  console.log(liquidFiles);
```
Woetfin, line 50: This console.log can be removed?

lib/utils/fsUtils.js
```diff
@@ -96,63 +124,89 @@ async function createLiquidFile(relativePath, fileName, textContent) {
...
+      break;
+    case "exportFile":
+      config.id = {};
+      config.name = `${handle}`;
+      config.file_name = "export_file.sxbrl";
+      config.text = "main.liquid";
+      config.encoding = "UTF-8";
+      config.text_parts = {};
+      break;
+    case "accountDetail":
```
Woetfin, line 191: Should be `accountTemplate`

lib/utils/fsUtils.js
```diff
@@ -2,88 +2,116 @@ const fs = require("fs");
 const path = require("path");
 const CWD = process.cwd();
 
+const FOLDERS = {
+  reconciliationText: "reconciliation_texts",
+  sharedPart: "shared_parts",
+  exportFile: "export_files",
+  accountDetail: "account_details",
```
Woetfin, line 9: Better that we refer to them as `accountTemplates`

lib/reconciliationText.js
```diff
@@ -0,0 +1,198 @@
...
+  }
+
+  /** Update template's id for the corresponding firm in template's config file */
+  static updateTemplateId(firmId, handle, templateId) {
+    let templateConfig = fsUtils.readConfig(this.TEMPLATE_TYPE, handle);
+    templateConfig.id[firmId] = templateId;
+    fsUtils.writeConfig(this.TEMPLATE_TYPE, handle, templateConfig);
+  }
+
+  static #missingHandle(template) {
```
Woetfin, line 102: Why exactly is the `#` in front of certain static functions?

.gitignore
```diff
@@ -9,3 +9,4 @@ shared_parts
 account_templates
 permanent_texts
 export_files
+account_details
```
Woetfin, line 12: Should be account_templates

Woetfin (changes requested):
Tested all commands:
* import-reconciliation: 💯
* update-reconciliation:
  * `.option("-a, --all", "Import all reconciliations")` --> mentions 'import'
* create-reconciliation: 💯
* import-shared-part: 💯
* update-shared-part: 💯
* create-shared-part: 💯
* add-shared-part: 💯
* remove-shared-part: 💯
* run-test: 
  * Running `run-test` always pops the message "Liquid testing file liquid_demo_liquid_test.yml already exists, so the file content was not overwritten" 
* authorize: 💯
* config: 💯
* get-reconciliation-id: 💯
* get-shared-part-id: 💯
* development-mode: 
  * with --test option it outputs some array to the console that doesn't seem needed: `development-mode --test sample_test`
* update: 
  * Maybe useful if we also add the command that's running? Just in case they still need to update it manually when the automatic update fails.

lib/reconciliationText.js
AgustinSilverfin: It will mark those functions as private. So they are only intended to be called from other functions inside the class.

lib/api/sfApi.js
AgustinSilverfin: I agree, but the need to use the `firmId` which can be provided via the CLI is the complex thing here. I guess we could modify it to set the firm id and do the check of the missing credentials somewhere else (directly by the cli as a first step).
But it sill be a big change, maybe we should tackle in a separate PR after.

Woetfin (commented):
Newest changes looking good! 
I added some changes for the update command as well, so the output is more detailed. 
Cause I sometimes got stuck / didn't complete running the command on WSL and had to go back to the docs to update manually. 

* update-reconciliation 💯
* run-test 💯
* development-mode 💯

lib/cli/cliUpdates.js
```diff
@@ -42,24 +45,30 @@ async function checkVersions() {
...
+    return;
+  } else {
+    console.log(`--------------`);
+    console.log(updateOutput.stdout);
+    console.log(
+      chalk.bold(`Silverfin CLI updated to version ${updatedPkgVersion.stdout}`)
+    );
+  }
+
+  return updateOutput;
```
AgustinSilverfin, line 71: Error are not catched correctly.

Before:

Screenshot from 2023-09-06 15-13-49

After update:

Screenshot from 2023-09-06 15-13-31

Woetfin: Added try/catch block so error fails gracefully

Woetfin (approved):

PR #57, opened 2023-08-29
Message flag

### LINKED ISSUE #51: Templates: add message to update-reconciliation command
When we use update-reconciliation, we are currently not able to add a description to the update. 
It always says **Update published using the API**
Screenshot 2023-08-10 at 09 23 21

It would be usefull for testing in the platform that we can add a description to the updates we are pushing to the platform
`silverfin update-reconciliation -h my_template -f 12345 -m "beta release order 67"`

AgustinSilverfin:
## Description

Support message flag

Fixes #51 

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -39,7 +39,11 @@ async function fetchAllReconciliations(firmId, page = 1) {
   fetchAllReconciliations(firmId, page + 1);
 }
 
-async function publishReconciliationByHandle(firmId, handle) {
+async function publishReconciliationByHandle(
+  firmId,
+  handle,
+  message = "Updated through the API"
+) {
   try {
     const templateConfig = fsUtils.readConfig("reconciliationText", handle);
     if (!templateConfig || !templateConfig.id[firmId]) {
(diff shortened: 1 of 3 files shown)


index.js
```diff
@@ -81,7 +92,7 @@ async function newReconciliation(firmId, handle) {
       return;
     }
     const template = await ReconciliationText.read(handle);
-    template.version_comment = "Updated through the CLI";
+    template.version_comment = message;
```
Toby-Masters-SF, line 95: [TO CONSIDER] Just to confirm, but would this message be visible anywhere as I checked the changelog and it seems only the update messages are visible? (However wouldn't do any harm to retain this code just in case)
6A81BA1E-650A-489A-B84A-2FE00AD7E29F

Toby-Masters-SF (approved):
All looks good. Reviewed code and functionally tested. Only point to possibly consider is that when creating reconciliations, no message is available on the Platform Changelog

index.js
AgustinSilverfin: Good point, we don't need the message while creating :slightly_smiling_face:

PR #58, opened 2023-08-30
Store firm name

AgustinSilverfin:
## Description

When we perform an authorization, we will now start fetching and storing the firm name as well.
We can later use to indicate more clearly with which firm we are interacting by displaying the name instead of (or next to) the id.
Specially useful for the Extension panel where we now show a list of firm ids, we can start displaying the firm name as well which is much more meaningful.

config should look like now:

```json
{
  "1234": {
    "accessToken": "...",
    "refreshToken": "...",
    "firmName": "My demo firm name"
  },
}
```

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist


Woetfin (approved):
Looks good!

PR #59, opened 2023-08-30
Verbose flag

### LINKED ISSUE #37: CLI: reducing printed information when using import-shared-part command

AgustinSilverfin:
## Description

Implement the use of package `consola`
Differentiate different output types `error`, `log`, `debug`, `info`,  `success`, `warn`
By default use a level of `info`.
Add the `--verbose` flag which will set the the level to `debug`.

Fixes #37 

~~BLOCKED BY: #56~~
~~BLOCKED BY: #57~~ 
~~BLOCKED BY: #50~~

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist


lib/utils/apiUtils.js
```diff
@@ -111,27 +109,27 @@ async function responseErrorHandler(
...
-    console.log(
+    consola.error(
       `Response Error (400): ${JSON.stringify(error.response.data.error)}`
     );
     return;
   }
   // No access credentials
   if (error.response.status === 401) {
-    console.log(
+    consola.error(
```
Woetfin, line 132: Also add this under `consola.debug` flag? It'll generally only trigger when refresh tokens are required. 

Unless the access token is completely invalid, we can have a separate error message to inform the user on that. 

But if code 401 + the error message is "The access token expired", we can skip any output in the console imo.

lib/utils/fsUtils.js
```diff
@@ -60,7 +72,7 @@ async function createLiquidTestYaml(templateType, handle, testContent) {
   );
   const existingFile = fs.existsSync(liquidTestPath);
   if (existingFile) {
-    console.log(
+    consola.debug(
```
Woetfin, line 75: Seems like it could be useful for a user as info? 
The user might not expect a liquid testing YAML to not be overwritten when importing.

lib/utils/fsUtils.js
```diff
@@ -69,7 +81,7 @@ async function createLiquidTestYaml(templateType, handle, testContent) {
     if (error) {
       errorCallbackLiquidTest(error);
     } else {
-      console.log(`Liquid testing YAML file created for ${handle}`);
+      consola.debug(`Liquid testing YAML file created for ${handle}`);
```
Woetfin, line 84: Could also still be useful as info imo

lib/utils/fsUtils.js
```diff
@@ -117,7 +129,7 @@ async function createTemplateFiles(templateType, handle, textMain, textParts) {
     );
     fs.writeFile(partPath, textParts[textPartName], emptyCallback);
   });
-  console.log(`Liquid template file(s) created for ${handle}`);
+  consola.debug(`Liquid template file(s) created for ${handle}`);
```
Woetfin, line 132: Could also still be useful as info imo

index.js
```diff
@@ -397,19 +541,26 @@ async function removeSharedPartFromReconciliation(
...
     case "sharedPart":
       templateText = await SF.findSharedPartByName(firmId, handle);
       break;
+    case "accountTemplate":
+      // To be implemented
+      break;
   }
   if (!templateText) {
-    console.log(`${handle} wasn't found`);
+    consola.debug(`Template ${handle} wasn't found (${type})`);
```
Woetfin, line 563: should be a consola.warn?

index.js
```diff
@@ -418,7 +569,7 @@ async function getTemplateId(firmId, type, handle) {
   }
   config.id[firmId] = templateText.id;
   fsUtils.writeConfig(type, handle, config);
-  console.log(`${handle}: ID updated`);
+  consola.debug(`Template ${handle}: ID updated (${type})`);
```
Woetfin, line 572: Should be consola.success?

Woetfin (changes requested):
Great work on the update & nice find with consola! 

Besides the code comments, also some general remarks: 

Some merge conflicts still outstanding

### Tested commands:
- run-test: OK
- import-reconciliation: OK
- update-reconciliation: OK
- get-reconciliation-id: No output of success

### consola.prompt for authorization flow

Consola also has a prompt function, for the authorization command, it wouldn't be too big of a refactor to already use this. 
Created separate issue to integrate this for the rest of the commands: https://github.com/silverfin/silverfin-cli/issues/83

```
consola.info(

      `NOTE: if you need to exit this process you can press "Ctrl/Cmmd + C"`

    );

    const redirectUri = "urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob";

    const scope =

      "administration%3Aread%20administration%3Awrite%20financials%3Aread%20financials%3Awrite%20workflows%3Aread";

  

    const firmIdPrompt = await consola.prompt("Enter the firm ID: ", {

      initial: firmId,

    });

  

    const url = `${apiUtils.BASE_URL}/f/${firmIdPrompt}/oauth/authorize?client_id=${process.env.SF_API_CLIENT_ID}&redirect_uri=${redirectUri}&response_type=code&scope=${scope}`;

  

    await open(url);

    consola.info(`You need to authorize your APP in the browser`);

    consola.info("Insert your credentials...");

    const authCodePrompt = await consola.prompt(

      "Enter your API authorization code: ",

      {

        echo: "*",

      }

    );
```

lib/cli/utils.js
```diff
@@ -39,7 +40,7 @@ function promptConfirmation() {
     "This will overwrite existing templates. Do you want to proceed? (y/n): "
```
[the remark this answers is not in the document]
AgustinSilverfin: Logged as a separater issue #83

lib/utils/fsUtils.js
AgustinSilverfin: In that case it will always be shown. By default we show info. I think once you know it will create those files, you would improve your experience reducing the terminal output (it won't add any value to you)

lib/utils/fsUtils.js
AgustinSilverfin: In that case it will always be shown. By default we show info. I think once you know it will create those files, you would improve your experience reducing the terminal output (it won't add any value to you)

lib/api/sfApi.js
```diff
@@ -26,20 +27,20 @@ async function authorizeApp(firmId = undefined) {
     const url = `${apiUtils.BASE_URL}/f/${firmIdPrompt}/oauth/authorize?client_id=${process.env.SF_API_CLIENT_ID}&redirect_uri=${redirectUri}&response_type=code&scope=${scope}`;
 
     await open(url);
-    console.log(`You need to authorize your APP in the browser`);
-    console.log("Insert your credentials...");
+    consola.info(`You need to authorize your APP in the browser`);
```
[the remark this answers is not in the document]
AgustinSilverfin: separate issue #83

lib/utils/fsUtils.js
AgustinSilverfin: In that case it will always be shown. By default we show info. I think once you know it will create those files, you would improve your experience reducing the terminal output (it won't add any value to you)

lib/utils/apiUtils.js
AgustinSilverfin: Changed

AgustinSilverfin:
Changes implemented. Also rebased (again) from export-files to align.

Woetfin (approved):
Looks good as well to go live!

PR #62, opened 2023-09-11
shared part main part is the same name as the shared part instead of 'main'

Woetfin:
## Description

The function readMainLiquid for a shared part was expecting the main part to be named 'main', but this should be the name of the shared part.

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/sharedPart.js
+++ b/lib/sharedPart.js
@@ -146,8 +146,8 @@ class SharedPart {
     }
   }
 
-  static #readMainLiquid(name, templateConfig) {
-    const mainPartPath = `./${this.TEMPLATE_FOLDER}/${name}/${templateConfig.text}`;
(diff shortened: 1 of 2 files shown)


lib/sharedPart.js
```diff
@@ -146,8 +146,8 @@ class SharedPart {
     }
   }
 
-  static #readMainLiquid(name, templateConfig) {
-    const mainPartPath = `./${this.TEMPLATE_FOLDER}/${name}/${templateConfig.text}`;
```
AgustinSilverfin, line 150: @Woetfin I think the issue is your `config.json` file, the `text` property.
In my opinion, we should use that property here rather calling the `{name}.liquid`

PR #69, opened 2023-09-18
fix status only run-test to take the testsRun object + skip spinner in this case

### LINKED ISSUE #68: Liquid testing: --status output returns FAILED while normal output mentions ALL TESTS PASSED
**Describe the bug**

--status output returns FAILED while normal output mentions ALL TESTS PASSED:

Screenshot 2023-09-15 at 17 22 18

**To Reproduce**
1. Go to this pull request: https://github.com/silverfin/be_market/pull/312
2. Run silverfin run-test -h vkt_1 --> All tests pass
3. Run silverfin run-test -h vkt_1 --status --> Mentions FAILED

**Expected behavior**
The output from --status should be the same pass/fail

 - OS: Mac
 - CLI version (run `silverfin --version`): [e.g. 1.18.1]

Woetfin:
## Description

The function for run-test with status only was not taking the updated object yet after the preview updates + skipped spinner in this case so the "Running tests" text isn't printed in the GitHub Actions output.

Fixes #68

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/liquidTestRunner.js
+++ b/lib/liquidTestRunner.js
@@ -80,12 +80,15 @@ function buildTestParams(firmId, handle, testName = "", renderMode) {
   return testParams;
 }
 
-async function fetchResult(firmId, testRunId) {
+async function fetchResult(firmId, testRunId, skipSpinner = false) {
   let testRun = { status: "started" };
   let pollingDelay = 1000;
   const waitingLimit = 500000;
 
-  spinner.spin("Running tests..");
+  if (!skipSpinner) {
+    spinner.spin("Running tests..");
+  }
+
   let waitingTime = 0;
   while (testRun.status === "started") {
     await new Promise((resolve) => {
@@ -343,7 +346,8 @@ async function runTests(
   handle,
   testName = "",
   previewOnly = false,
-  renderMode = "none"
+  renderMode = "none",
+  skipSpinner = false
 ) {
   try {
     const testParams = buildTestParams(firmId, handle, testName, renderMode);
@@ -361,7 +365,7 @@ async function runTests(
     if (!previewOnly) {
       const testRunResponse = await SF.createTestRun(firmId, testParams);
       const testRunId = testRunResponse.data;
-      testRun = await fetchResult(firmId, testRunId);
+      testRun = await fetchResult(firmId, testRunId, skipSpinner);
     }
 
     return { testRun, previewRun };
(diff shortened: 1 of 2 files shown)


lib/liquidTestRunner.js
```diff
@@ -409,9 +413,12 @@ async function runTestsWithOutput(
 // RETURN (AND LOG) ONLY PASSED OR FAILED
 // CAN BE USED BY GITHUB ACTIONS
 async function runTestsStatusOnly(firmId, handle, testName = "") {
-  const testRun = await runTests(firmId, handle, testName, false, "none");
-  if (testRun && testRun.status === "completed") {
-    const errorsPresent = checkAllTestsErrorsPresent(testRun.tests);
+  const testsRun = await runTests(firmId, handle, testName, false, "none", true);
```
AgustinSilverfin, line 416: @Woetfin why the extra `true`?
`runTests` has 5 parameters but here we are passing 6 arguments

lib/liquidTestRunner.js
```diff
@@ -343,7 +346,8 @@ async function runTests(
   handle,
   testName = "",
   previewOnly = false,
-  renderMode = "none"
+  renderMode = "none",
+  skipSpinner = false
```
AgustinSilverfin, line 350: Skipping the spinner is not really necessary. Is not the reason of this bug. 
But we could add it if you think it's useful.

lib/liquidTestRunner.js
```diff
@@ -409,9 +413,12 @@ async function runTestsWithOutput(
 // RETURN (AND LOG) ONLY PASSED OR FAILED
 // CAN BE USED BY GITHUB ACTIONS
 async function runTestsStatusOnly(firmId, handle, testName = "") {
-  const testRun = await runTests(firmId, handle, testName, false, "none");
```
AgustinSilverfin, line 412: We can actually only change this row to:

```javascript
  const testRuns = await runTests(firmId, handle, testName, false, "none");
  const testRun = testRuns.testRun;
```

And I think that should be all to work again

lib/liquidTestRunner.js
Woetfin: I'll remove it, cause it wouldn't have really solved the issue I wanted and would take a bigger code change. So would take it into separate PR

lib/liquidTestRunner.js
Woetfin: was for the skipSpinner

PR #76, opened 2023-10-09
Fix: template id and Partners

### LINKED ISSUE #74: CLI: development-mode gets into some kind of loop
**Describe the bug**

When activating the development mode with the command `silverfin development-mode --update-templates --yes`, a weird thing happens where all of a sudden several RT's are updated (which didn't have to do anything with the change I did in another RT). 

Screenshot of the updates being in a loop: 
Screenshot 2023-10-04 at 13 53 28

Screenshot of the changed files of unrelated RT's: 
Screenshot 2023-10-04 at 13 53 46

Screenshot of the done change: 
Screenshot 2023-10-04 at 13 53 51

**To Reproduce**
1. Activate development mode
2. Add change to local file 
3. Save the file 

**Expected behavior**

What I would expect, is in the terminal an update is done in the related RT of the change (and only that). 

See above

 - OS: MacOS Ventura 13.6
 - CLI version (run `silverfin --version`): **1.20.2**

**Additional context**

Was discussed with Woet as well; see [here](https://silverfin.slack.com/archives/C0376K9NNTS/p1695906264911449)

AgustinSilverfin:
## Description

- Updating templates to Silverfin: fetch the result and log if it was updated or not.
- Modify the way we look for reconciliations. We skip a template if it's from Partners (before we relied on the visible text but that could be visible from partners, so now we rely on the marketplace_template_id present or not -> templates that have this value are certainly linked to partners). We still fetch the first non-partners template, so it's not possible to have multiple templates with the same handle (except partners) in the environment.
- Move missing template id errors to `errorUtils`

Fixes #74 

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/api/sfApi.js
+++ b/lib/api/sfApi.js
@@ -123,25 +123,44 @@ async function readReconciliationTextById(firmId, id, refreshToken = true) {
   }
 }
 
+/**
+ * Find reconciliation by handle. We only look for templates that are not from Partners (because tecnically they don't belong to the current firm). If we find a template from Partners we will skip it and continue looking for the next one (while we keep it's `marketplace_template_id` as a reference)
+ * @param {Number} firmId
+ * @param {String} handle
+ * @param {Number} page
+ * @returns {Object} Reconciliation Text object
+ */
 async function findReconciliationTextByHandle(firmId, handle, page = 1) {
+  let marketplace_template_id;
   const reconciliations = await readReconciliationTexts(firmId, page);
-  // No data
+  // No data. Not found
   if (reconciliations.length == 0) {
     console.log(`Reconciliation ${handle} not found`);
-    return;
+    return null;
   }
   let reconciliationTexts = reconciliations.filter(
     (element) => element["handle"] === handle
   );
-  if (reconciliationTexts.length != 0) {
-    for (let reconciliationText of reconciliationTexts) {
-      // Only return reconciliations were liquid code is not hidden
-      if (reconciliationText.hasOwnProperty("text")) {
-        return reconciliationText;
-      }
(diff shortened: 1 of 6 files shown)


lib/api/sfApi.js
```diff
@@ -123,25 +123,44 @@ async function readReconciliationTextById(firmId, id, refreshToken = true) {
...
+ * @param {Number} page
+ * @returns {Object} Reconciliation Text object
+ */
 async function findReconciliationTextByHandle(firmId, handle, page = 1) {
+  let marketplace_template_id;
   const reconciliations = await readReconciliationTexts(firmId, page);
-  // No data
+  // No data. Not found
   if (reconciliations.length == 0) {
     console.log(`Reconciliation ${handle} not found`);
```
Woetfin, line 138: This line seems to be never reached

AgustinSilverfin: It is reach when we call a page with no more reconciliations in it. Basically, when the handle provided doesn't exists

Woetfin: ah ok, I see in that case it is reached, I was testing the case where you're trying to import a marketplace recon, so it does exist, but it also doesn't import anything. So it should also give a message that it wasn't imported, but it silently doesn't do anything. 

For example this one: 
`bin/cli.js import-reconciliation --yes -f 542 -h vol_6_19`

Woetfin (approved):
Changes looks good! 👍

lib/api/sfApi.js
AgustinSilverfin: As discussed, I have removed `marketplace_template_id`. We will consider the best flow while developing the Partners API

PR #80, opened 2023-10-16
Support managing account templates

AgustinSilverfin:
## Description

Support account template in respository. Create, Read and Update operations.

TO-DO: decide which is going to be their name identifier, since account templates don't have a `handle` like reconciliations or a `name` that is present in every locale.

~~Blocked by #50~~ 
~~Blocked by #59~~

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist


Woetfin (approved):
@AgustinSilverfin updated & tested all the commands for managing the account templates:

* import-account-template: name, id, all
* update-shared-part: name, all
* create-shared-part: name
* add-shared-part: name
* remove-shared-part: name

Can you review if this looks ok for you?

PR #81, opened 2023-10-18
Make tests pass on an empty liquid testing YAML

Woetfin:
## Description

The run-tests with --status were passing "FAILED" on an empty liquid testing YAML, but this caused actions to crash. 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/liquidTestRunner.js
+++ b/lib/liquidTestRunner.js
@@ -32,7 +32,7 @@ function buildTestParams(firmId, handle, testName = "", renderMode) {
   const relativePath = `./reconciliation_texts/${handle}`;
   const config = fsUtils.readConfig("reconciliationText", handle);
   const testPath = `${relativePath}/${config.test}`;
-  const testContent = fs.readFileSync(testPath, "utf-8");
+  const testContent = fs.readFileSync(testPath, "utf-8").trim();
 
   // Empty YAML check
   if (testContent.split("\n").length <= 1) {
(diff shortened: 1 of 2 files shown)


lib/liquidTestRunner.js
```diff
@@ -409,17 +409,27 @@ async function runTestsWithOutput(
 // RETURN (AND LOG) ONLY PASSED OR FAILED
 // CAN BE USED BY GITHUB ACTIONS
 async function runTestsStatusOnly(firmId, handle, testName = "") {
+  let status = "FAILED";
   const testResult = await runTests(firmId, handle, testName, false, "none");
```
AgustinSilverfin, line 413: If the YAML file has only one empty row, then it will be `undefined` as you expect.
But, if the YAML has two or more empty rows, then it will no longer be undefined, returning a FAILED.

Example with two empty rows :thinking: 
```
{
  testRun: {
    status: 'test_error',
    error_message: 'Test was not found. Please make sure to put the cursor on your test definition line',
    error_line_number: 1
  },
  previewRun: null
}
```

AgustinSilverfin (changes requested):

lib/liquidTestRunner.js
Woetfin: Nice find! I've added a trim() to the so if there are only spaces or newlines, it removes them and the testContext remains empty. This error might still occur though if all you have are comments in the file.

AgustinSilverfin (approved):

PR #82, opened 2023-10-19
Prioritize id over handle or name

AgustinSilverfin:
## Description

When we import a template, we usually use it's handle or name because we don't have the ID, or because fetching the ID is more time-consuming for the user. We use these handle to identify which is the template, from all the existing ones, that we need to fetch.

But, once the template already exists (it is in the repo and has an existing config file), we may already have it's corresponding ID stored there. If, by any reason, we would like to re-import it's template, there is no need to use the handle since we can already use the previously stored ID.

TL;DR: If we already have in the config, use the ID. If not, rely on the handle/name.

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist


index.js
```diff
@@ -7,6 +7,15 @@ const { ReconciliationText } = require("./lib/reconciliationText");
 const { SharedPart } = require("./lib/sharedPart");
 const { firmCredentials } = require("./lib/api/firmCredentials");
 
+async function fetchReconciliation(firmId, handle) {
+  const templateConfig = fsUtils.readConfig("reconciliationText", handle);
+  if (templateConfig || templateConfig.id[firmId]) {
```
Woetfin, line 12: I think we should do an "&&" check instead of "||"? 

Cause templateConfig could be true if the config.json was already created from importing from another firm and so this if-statement would pass, but the ID from the specific firm you want to import from doesn't exist yet.

index.js
```diff
@@ -111,21 +120,30 @@ async function newAllReconciliations(firmId) {
...
 }
 
-async function importExistingSharedPartById(firmId, id) {
-  const sharedPart = await SF.readSharedPartById(firmId, id);
-
-  if (!sharedPart) {
-    throw `Shared part ${id} wasn't found.`;
+async function fetchSharedPart(firmId, name) {
+  const templateConfig = fsUtils.readConfig("sharedPart", name);
+  if (templateConfig || templateConfig.id[firmId]) {
```
Woetfin, line 125: I think the same applies here as well.

Woetfin (changes requested):
I think it's a good idea to prep this update, cause it's a problem I've also encountered when importing and this would make it more intuitive 👍Do still have an open comment on the condition in the if-statement.

index.js
AgustinSilverfin: Yes, you are right! :see_no_evil:

PR #84, opened 2023-10-31
Fix mkdir so complete path is always created + show success message on shared part import

Woetfin:
## Description

1. The create/import was throwing an error if the parent reconciliation_texts or shared_parts folder didn't exist yet. 

2. Import a shared part didn't give any success messages

## Type of change

- [x] Bug fix
- ~~New feature~~
- ~~Breaking change~~

## Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -249,6 +249,7 @@ async function fetchSharedPart(firmId, name) {
   } else {
     fetchSharedPartByName(firmId, name);
   }
+  consola.success(`Shared part "${templateConfig.name}" imported`);
 }
 
 async function fetchSharedPartById(firmId, id) {
@@ -258,7 +259,6 @@ async function fetchSharedPartById(firmId, id) {
     process.exit(1);
   }
   await SharedPart.save(firmId, template.data);
-  consola.success(`Shared part "${template.data.name}" imported`);
 }
 
 async function fetchSharedPartByName(firmId, name) {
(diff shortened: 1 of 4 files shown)


index.js
```diff
@@ -249,6 +249,7 @@ async function fetchSharedPart(firmId, name) {
   } else {
     fetchSharedPartByName(firmId, name);
   }
+  consola.success(`Shared part "${templateConfig.name}" imported`);
```
AgustinSilverfin, line 252: @Woetfin why did you move this here? now is no longer consistent with `fetchReconciliation` for example.

lib/templates/sharedPart.js
```diff
@@ -97,11 +97,7 @@ class SharedPart {
 
   static async #findTemplateHandle(firmId, template) {
     // Search in repository
-    let handle = await fsUtils.findHandleByID(
-      firmId,
-      template.type,
-      template.id
-    );
+    let handle = fsUtils.findHandleByID(firmId, template.type, template.id);
```
AgustinSilverfin, line 100: @Woetfin any specific reason to remove the `await` ?

index.js
WoetDev: For some reason the success message wasn't logging on the previous place, but it was logging here. Reduces duplicate code though so I think it's fine to leave it for now (need to go over all the functions either way for the partner updates, so we can align)

lib/templates/sharedPart.js
WoetDev: VS Code let me auto-remove it because they said it's unnecessary, so I followed that advice

PR #90, opened 2023-11-21
Check name stored in config

### LINKED ISSUE #88: CLI: Check when template name has changed
**Describe the bug**

Currently, we use handle, name or name_nl (depending on the template type) to name the template's directory, and also to identify templates from commands.

If, for any reason, a template is renamed, the corresponding config.json file will reflect the new name while the directory will still be named with it's original name. Which implies that commands should be still run by using it's old name.

**To Reproduce**

1. Create a reconciliation using `create-reconciliation --handle original_handle`
2. Go to the reconciliations `config.json` and update it's `handle` to `new_handle`
3. Update the reconciliation with `update-reconciliation --handle original_handle`
4. Go to the Platform to see the template's code to confirm that handle has changed.
5. Try to fetch it using `import-reconciliation --handle new_handle` (which should fail)
6. Fetch the reconciliation using the old handle `import-reconciliation --handle original_handle` 

**Proposed Solution**

Create an extra check that warn users when the handle/name/name_nl attribute from the config differs from the template's directory name

**Expected behavior**

If `handle/name/name_nl` of a template change, you should be able to run commands with it's new name.

**Additional context**

This should be address for every client type.

AgustinSilverfin:
## Description

- Check if the name stored in the config is still aligned with the one used to run commands and name directories
- Avoid creating new directories and config if the name/handle used in import commands is related to a template that doesn't exist

Fixes #88 

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/utils/templateUtils.js
+++ b/lib/utils/templateUtils.js
@@ -79,4 +79,5 @@ module.exports = {
   checkValidName,
   filterParts,
   missingLiquidCode,
+  getTemplateName,
 };
--- a/index.js
+++ b/index.js
@@ -28,7 +28,7 @@ async function fetchReconciliationByHandle(firmId, handle) {
     consola.error(`Reconciliation "${handle}" wasn't found`);
     process.exit(1);
   }
-  const saved = ReconciliationText.save(firmId, template);
+  const saved = await ReconciliationText.save(firmId, template);
   if (saved) {
     consola.success(`Reconciliation "${handle}" imported`);
   }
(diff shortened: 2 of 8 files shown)


lib/utils/templateUtils.js
```diff
@@ -15,7 +27,7 @@ function checkValidName(name, templateType) {
         templateType === "accountTemplate" || templateType === "exportFile"
           ? `Valid ${templateType} names can't include back- or forward slashes`
           : `Valid ${templateType} names only include alphanumeric characters and underscores`
-      }.`
+      }.`,
```
Woetfin, line 30: unneeded comma?

lib/utils/templateUtils.js
```diff
@@ -36,7 +48,7 @@ function missingLiquidCode(template) {
     consola.warn(
       `Template "${
         template?.handle || template?.name || template?.name_nl
-      }": this template's liquid code was empty or hidden so it was not imported.`
+      }": this template's liquid code was empty or hidden so it was not imported.`,
```
Woetfin, line 51: unneeded comma?

Woetfin (approved):
Some small comments but update looks good

PR #92, opened 2023-11-29
Remove logic to immediately create a config.json before we know the command is running for an existing template

Woetfin:
## Description

Removed the logic to immediately create a config.json if it's missing since this triggers a new directory & json to already be created before we know the template they're trying to interact with exists. The config.json creation should already be covered now in the save() functions in the template classes so I believe it can be safely removed. 

## Type of change

- [x] Bug fix
- ~~New feature~~
- ~~Breaking change~~

## Checklist

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "silverfin-cli",
-  "version": "1.24.2",
+  "version": "1.24.3",
   "description": "Command line tool for Silverfin template development",
   "main": "index.js",
   "license": "MIT",


lib/utils/fsUtils.js
```diff
@@ -156,73 +156,23 @@ function configExists(templateType, handle) {
...
-}
+  try {
+    const configPath = path.join(
+      process.cwd(),
+      `${FOLDERS[templateType]}`,
+      `${handle}`,
+      "config.json"
+    );
 
-/**
```
AgustinSilverfin, line 170: Not sure why we are removing the entire creation of config, this will not only affect the reading of configs

PR #100, opened 2024-01-18
Avoid errors to terminate process

AgustinSilverfin:
## Description

Right now, most error handling functions raise and error and terminate the process with `process.exit(1)`. 
Changes this behavior to display the error but continue, only returning false. Allowing to the caller to handle the exception as needed.
This could be used in the Extension for example, or in commands like `update-reconciliation --all` where any error would stop the execution. 

## Type of change

- [x] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


Woetfin (approved):

Woetfin (approved):
Fix looks good!

PR #130, opened 2024-05-17
Feature: support running liquid tests for account templates

Woetfin:
## Description

Support running liquid tests for account templates

Some example templates to test you can find here: [account_templates.zip](https://github.com/silverfin/silverfin-cli/files/15414151/account_templates.zip)

The tests run, but there seem to be two issues that at first glance might be core issues that ENG needs to solve:

1. Difference in reconciled expectation output on platform compared to API

When I run the tests on this template on the platform: https://live.getsilverfin.com/f/14400/account_detail_templates/492722/edit

All the tests pass on the platform. 
Screenshot 2024-05-23 at 10 43 20

But running that exact same test suite with the API, results in a different reconciled expectation output:
Screenshot 2024-05-23 at 10 44 20

2. The HTML seems to always return the export view of the template, even if mode "input" or "all" is specified:

Running the test with only expected input view rendered for the template:
Screenshot 2024-05-23 at 10 47 04

Still results in the export view being returned by the API:
Screenshot 2024-05-23 at 10 30 35

## Type of change

- ~~Bug fix~~
- [x] New feature
- ~~Breaking change~~

## Checklist


AgustinSilverfin (commented):
In general looks good, my comments are mostly about the style we follow and decision to be explicit or implicit for what we are introducing and the selection of template types.

I think being implicit in this case could be more error prone.

I'm glad that not too many changes are required :smile:

bin/cli.js
```diff
@@ -391,23 +391,37 @@ program
   .command("run-test")
   .description("Run Liquid Tests for a reconciliation template from a YAML file")
   .requiredOption("-f, --firm <firm-id>", "Specify the firm to be used", firmIdDefault)
-  .requiredOption("-h, --handle <handle>", "Specify the reconciliation to be used (mandatory)")
+  .option("-h, --handle <handle>", "Specify the reconciliation to be used (mandatory)")
+  .option("-at, --account-template <name>", "Specify the reconciliation to be used (mandatory)")
```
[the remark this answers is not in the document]
michieldegezelle: Adjusted!

lib/liquidTestRunner.js
```diff
@@ -29,18 +31,28 @@ function findTestRows(testContent) {
   return indexes;
 }
 
-function buildTestParams(firmId, handle, testName = "", renderMode) {
-  const relativePath = `./reconciliation_texts/${handle}`;
+function buildTestParams(
+  firmId,
+  templateType = "reconciliationText",
```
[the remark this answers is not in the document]
michieldegezelle: Good point! See my response in your other comment

lib/api/sfApi.js
```diff
@@ -401,9 +401,18 @@ async function removeSharedPartFromAccountTemplate(type, envId, sharedPartId, ac
...
+      const response = await instance.post(
+        "account_templates/test",
+        attributes
+      );
+
+      return response;
+    }
+
     const response = await instance.post("reconciliations/test", attributes);
     return response;
```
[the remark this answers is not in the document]
michieldegezelle: Another question:

Is it fine to have a try/catch and a switch (with default and error handling) combined? Or is this overkill?

bin/cli.js
```diff
@@ -391,23 +391,31 @@ program
...
   .option("-t, --test <test-name>", "Specify the name of the test to be run (optional)", "")
   .option("--html-input", "Get a static html of the input-view of the template generated with the Liquid Test data (optional)", false)
   .option("--html-preview", "Get a static html of the export-view of the template generated with the Liquid Test data (optional)", false)
   .option("--preview-only", "Skip the checking of the results of the Liquid Test in case you only want to generate a preview template (optional)", false)
   .option("--status", "Only return the status of the test runs as PASSED/FAILED (optional)", false)
+
   .action((options) => {
-    cliUtils.checkDefaultFirm(options.firm, firmIdDefault);
+    if (!options.handle && !options.accountTemplate) {
+      consola.error("You need to specify either a handle or an account template");
```
AgustinSilverfin, line 404: Nitpick: "reconciliation handle"?

lib/liquidTestRunner.js
AgustinSilverfin, line 1: We are modifying in  `runTests`,   `runTestsWithOutput`,  `runTestsStatusOnly` the amount of parameters and their order. This functions are exported, can be called externally (even from the VSCode extension is mentioned)
Considering that, I will assert at the very beginning of each of them, that `templateType` is properly passed as `accountTemplate` or `reconciliationText`. I think that could help us identify and avoid some hard to debug issues otherwise

lib/cli/devMode.js
```diff
@@ -12,11 +12,29 @@ const chokidar = require("chokidar");
...
+      consola.info(`Watching for changes related to reconciliation "${handle}" to run a new test...`);
+      filePath = path.resolve(process.cwd(), "reconciliation_texts", handle, "tests", `${handle}_liquid_test.yml`);
+      break;
+    case "accountTemplate":
+      consola.info(`Watching for changes related to account template "${handle}" to run a new test...`);
+      filePath = path.resolve(process.cwd(), "account_templates", handle, "tests", `${handle}_liquid_test.yml`);
+      break;
+    default:
+      break;
+  }
```
AgustinSilverfin, line 35: This could be taken from fsUtils actually, right?

something like `fsUtils.FOLDERS[templateType]` may work :thinking:

lib/utils/fsUtils.js
```diff
@@ -300,17 +300,22 @@ function listExistingFiles(typeCheck = "liquid") {
...
+  if (templateType !== "reconciliationText" && templateType !== "accountTemplate") {
+    consola.error(`Template type is missing or invalid`);
+    process.exit(1);
+  }
+
+  const relatedSharedParts = listSharedPartsUsedInTemplate(firmId, templateType, handle);
   const allLiquidFiles = listExistingFiles("liquid");
-  const patternReconciliation = `reconciliation_texts/${handle}/`;
-  const reconciliationRegExp = new RegExp(patternReconciliation, "g");
+  const folderPath = templateType === "reconciliationText" ? `reconciliation_texts/${handle}/` : `account_templates/${handle}/`;
```
AgustinSilverfin, line 314: same here for the folder path, we already have `FOLDERS` we can use

PR #141, opened 2024-09-18
remove https from url in getAccessToken

JVersee:
## Description

BASE_URL const already includes https:// in `https://live.getsilverfin.com'.
In the config we have to use `https://` in that case as well and ensure no duplicate `https://` is being used when string interpolating.

## Type of change

- [x ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist


Toby-Masters-SF (approved):
Approved

PR #144, opened 2024-10-03
[Stats] Add percentage columns

AgustinSilverfin:
## Description

Calculate the amount of templates that are externally managed and have unit tests 
Add them to new columns in the output csv file
Include percentage in the overview displayed on the terminal

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/cli/stats.js
+++ b/lib/cli/stats.js
@@ -159,40 +159,51 @@ async function setMainPath(templateType, templateName) {
   return filePath;
 }
 
+function percentageRoundTwo(numerator, denominator) {
+  return denominator > 0 ? Number(((numerator / denominator) * 100).toFixed(2)) : 0;
+}
+
 async function getTemplatesSummary() {
   const summary = {
     reconciliations: {
       total: 0,
       externallyManaged: 0,
+      externallyManagedPerc: 0,
       yamlFiles: 0,
+      yamlFilesPerc: 0,
       unitTests: 0,
     },
     sharedParts: {
       total: 0,
       externallyManaged: 0,
+      externallyManagedPerc: 0,
     },
     exportFiles: {
       total: 0,
       externallyManaged: 0,
+      externallyManagedPerc: 0,
     },
     accountTemplates: {
       total: 0,
       externallyManaged: 0,
+      externallyManagedPerc: 0,
       yamlFiles: 0,
+      yamlFilesPerc: 0,
       unitTests: 0,
     },
     all: {
       total: 0,
       externallyManaged: 0,
+      externallyManagedPerc: 0,
       yamlFiles: 0,
+      yamlFilesPerc: 0,
       unitTests: 0,
     },
   };
 
   // Reconciliations
-  const reconciliationsNonEmpty = await listNonEmptyTemplates(
-    "reconciliationText"
-  );
+  const reconciliationsNonEmpty =
+    await listNonEmptyTemplates("reconciliationText");
   const reconciliationsExtMan = await listExternallyManagedTemplates(
     "reconciliationText",
     reconciliationsNonEmpty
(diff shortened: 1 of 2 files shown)


lib/cli/stats.js
```diff
@@ -159,48 +159,67 @@ async function setMainPath(templateType, templateName) {
...
+function percentageRoundTwo(num, den) {
+  return den > 0 ? Number(((num / den) * 100).toFixed(2)) : 0;
+}
+
 async function getTemplatesSummary() {
   const summary = {
     reconciliations: {
       total: 0,
       externallyManaged: 0,
+      externallyManagedPerc: 0,
```
Toby-Masters-SF, line 171: [COMMENT] Update summary structure in getTemplatesSummary for new % field

lib/cli/stats.js
```diff
@@ -159,48 +159,67 @@ async function setMainPath(templateType, templateName) {
...
+      yamlFilesPerc: 0,
       unitTests: 0,
     },
   };
 
   // Reconciliations
-  const reconciliationsNonEmpty = await listNonEmptyTemplates(
-    "reconciliationText"
-  );
+  const reconciliationsNonEmpty =
```
Toby-Masters-SF, line 205: [SUGGESTION] Restore assignment to 1 line for consistency with other assign statements?

lib/cli/stats.js
```diff
@@ -159,48 +159,67 @@ async function setMainPath(templateType, templateName) {
   return filePath;
 }
 
+function percentageRoundTwo(num, den) {
```
Toby-Masters-SF, line 162: [SUGGESTION] Worth renaming to 'numerator' and 'denominator' to make function more readable?

lib/cli/stats.js
```diff
@@ -159,48 +159,67 @@ async function setMainPath(templateType, templateName) {
...
+  return den > 0 ? Number(((num / den) * 100).toFixed(2)) : 0;
```
Toby-Masters-SF, line 163: [COMMENT] 
Cutoff String value at 2 dp. Agree don't need to be too precious about rounding as for reporting, not exact accounting

Nice safety check with den > 0 to avoid dividing by 0.

lib/cli/stats.js
```diff
@@ -220,19 +243,30 @@ async function getTemplatesSummary() {
...
+  summary.exportFiles.externallyManagedPerc = percentageRoundTwo(
+    summary.exportFiles.externallyManaged,
+    summary.exportFiles.total
+  );
 
   // Account Templates
-  const accountTemplatesNonEmpty = await listNonEmptyTemplates(
-    "accountTemplate"
-  );
+  const accountTemplatesNonEmpty =
```
Toby-Masters-SF, line 252: [SUGGESTION] Could further reduce to 1 line?

lib/cli/stats.js
```diff
@@ -246,8 +280,16 @@ async function getTemplatesSummary() {
...
     summary.accountTemplates.externallyManaged;
+  summary.all.externallyManagedPerc = percentageRoundTwo(
+    summary.all.externallyManaged,
+    summary.all.total
+  )
   summary.all.yamlFiles =
     summary.reconciliations.yamlFiles + summary.accountTemplates.yamlFiles;
+  summary.all.yamlFilesPerc = percentageRoundTwo(
+    summary.all.yamlFiles,
+    summary.all.total
```
Toby-Masters-SF, line 291: [TODO] Should we exclude shared parts (and possibly export files) from this total figure as _cannot_ have YAML tests for these items. Could also mean we need a separate total 'testable' value

Toby-Masters-SF (changes requested):
Hi Rufo, 

Review complete. Had a couple of very minor style suggestions (please feel free to take them or leave them)

Only point I wanted to push back on was the calculation of the total YAML test % - as I'm not sure whether we should include shared parts and export files

lib/cli/stats.js
AgustinSilverfin: Multi-line in this assignment was automatically implemented by the formatter (prettier). I  tend to respect them if not we will need to manually adjust them constantly.

FYI We could consider modifying the configuration of the formatter as we want, so far we only have two rules defined https://github.com/silverfin/silverfin-cli/blob/7d9ce941a15aa361753c59e6afba713a50a79996/package.json#L39

lib/cli/stats.js
AgustinSilverfin: Same as https://github.com/silverfin/silverfin-cli/pull/144#discussion_r1818930488

lib/cli/stats.js
AgustinSilverfin: Good catch. Updated to only consider RT and ADT

lib/cli/stats.js
AgustinSilverfin: Good catch. Updated to only consider RT and ADT

lib/cli/stats.js
Toby-Masters-SF: Happy to respect the formatter

Toby-Masters-SF (approved):
LGTM

PR #148, opened 2024-10-31
Support Staging environments

### LINKED ISSUE #131: CLI: support use of stagings
At the moment, the CLI can be used to interact with templates on the production Silverfin environment. But being able to change the host to some other than live.getsilverfin.com, should allow the interaction with a staging environment as well. 

Be able to change the host and interact with a staging from the CLI.

Still manual interactions on the platform needed when interacting with a staging.

**Additional context**
Setting a template to externally_managed on the live environment, causes this to be externally_managed on a staging as well. But at the moment you then can't interact through the CLI with these templates.

AgustinSilverfin:
## Description

Fixes #131 

For Silverfin Staging, Basic authentication needs to be added in the request header, with the proper credentials.

## Updates 

- New commands to set and get HOST
- Maintain also HOST as ENV variable for compatibility
- Support Staging Environments
- Log stored firms and host before each command
- Extract into dedicated class and add tests

## How to test

You need to create a new ENV variable named `SF_BASIC_AUTH`  (you can reach out to me in case you don't have it yet)

Different commands that should be tested.
NOTE: when you switch hosts (for example, from live to staging) a new `silverfin authorize` would need to be run first.
TIP: always use `--verbose` flag in all commands

```bash
silverfin authorize
silverfin config --get-host
silverfin config --set-host # switch between live and staging
silverfin import-reconciliation # from live/staging and firm/partners
silverfin config --refresh-token
silverfin config --refresh-partner-token
```

For example, some steps that can be followed:

```bash
silverfin --set-host <staging-url>
silverfin authorize
silverfin import-reconciliation # from firm
silverfin import-reconciliation # from partners
silverfin config --get-host
silverfin config --refresh-token
silverfin config --refresh-partner-token
silverfin --set-host <production-url>
silverfin authorize
silverfin import-reconciliation # from firm
silverfin import-reconciliation # from partners
silverfin config --refresh-token
silverfin config --refresh-partner-token
```

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -1028,5 +1028,6 @@ if (pkg.repository && pkg.repository.url) {
 (async function () {
   // Check if there is a new version available
   await cliUpdates.checkVersions();
+  cliUtils.logCurrentEnvironments();
   await program.parseAsync();
 })();
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -822,10 +822,7 @@ program
     }
     if (options.refreshToken) {
       cliUtils.checkDefaultFirm(options.refreshToken, firmIdDefault);
-      const refreshedTokens = await SF.refreshTokens(
-        "firm",
-        options.refreshToken
-      );
+      const refreshedTokens = await SF.refreshFirmTokens(options.refreshToken);
 
       if (refreshedTokens) {
         consola.success(
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -300,8 +300,8 @@ program
   .option("--yes", "Skip the prompt confirmation (optional)")
   .action((options) => {
     const settings = runCommandChecks(
-      ["name", "all"], 
-      options, 
+      ["name", "all"],
+      options,
       firmIdDefault,
       true,
       true // Message required
@@ -760,6 +760,15 @@ program
       "Get a new partner api key using the stored api key"
     )
   )
+  .addOption(
+    new Option(
+      "--set-host [host]",
+      "Set a custom host for the Silverfin API (e.g. https://live.getsilverfin.com)"
+    )
+  )
+  .addOption(
+    new Option("--get-host", "Get the current host for the Silverfin API")
+  )
   .action(async (options) => {
     cliUtils.checkUniqueOption(
       [
@@ -769,6 +778,8 @@ program
         "updateName",
         "refreshToken",
         "refreshPartnerToken",
+        "setHost",
+        "getHost",
       ],
       options
     );
@@ -832,6 +843,14 @@ program
           `Partner API key refreshed for partner ID: ${refreshedTokens.partner_id}`
         );
       }
+    }
+    if (options.setHost) {
+      firmCredentials.setHost(options.setHost);
+      consola.success(`Host set to: ${options.setHost}`);
+    }
+    if (options.getHost) {
+      const host = firmCredentials.getHost();
+      consola.info(`Current host: ${host}`);
     }
   });
 
--- a/lib/api/axiosFactory.js
+++ b/lib/api/axiosFactory.js
@@ -42,12 +42,17 @@ class AxiosFactory {
       process.exit(1);
     }
 
+    let baseHeaders = {
+      "User-Agent": `silverfin-cli/${pkg.version}`,
+      "X-Firm-ID": envId,
+    };
+
     let axiosDetails;
     if (this.#isStaging()) {
       axiosDetails = {
         baseURL: `${this.BASE_URL}/api/v4/f/${envId}`,
         headers: {
-          "User-Agent": `silverfin-cli/${pkg.version}`,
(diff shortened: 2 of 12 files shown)


lib/api/axiosFactory.js
```diff
@@ -0,0 +1,274 @@
...
+   * Get a new set of tokens for a given firm id
+   */
+  static async #refreshFirmTokens(axiosInstance, firmId) {
+    try {
+      consola.debug(`Refreshing tokens for firm ${firmId}`);
+      const firmTokens = firmCredentials.getTokenPair(firmId);
+      let data = {
+        client_id: process.env.SF_API_CLIENT_ID,
+        client_secret: process.env.SF_API_SECRET,
+        redirect_uri: "urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob",
```
agon-qurdina, line 240: Is this url encoded? Can we add the raw path instead and then run something like `.urlEncode()` (not sure this exists but you get the idea) for better readability?

lib/api/axiosFactory.js
```diff
@@ -0,0 +1,274 @@
...
+
+  /**
+   * Create an axios instance for a given type and environment id.
+   * It will add the necessary headers and interceptors to handle the authorization, and refresh the tokens if needed.
+   * The created instance will be used to make requests to the Silverfin API.
+   * @param {String} type - The type of instance to create (firm or partner)
+   * @param {Number} envId - The environment id to create the instance for
+   * @returns {Object} - The created axios instance
+   */
+  static createInstance(type, envId) {
```
agon-qurdina, line 17: Would it make sense to store these instances in memory iso creating a new one on every call? This would then become more of something like a `findOrCreateInstance` or `loadInstance`:

```
const key = `${type}_${envId}`
axios_instances[key] ||= this.#createAxiosInstanceForFirms(envId)
return axios_instances[key]
```

lib/api/axiosFactory.js
```diff
@@ -0,0 +1,274 @@
...
+        params: {
+          access_token: firmTokens.accessToken,
+        },
+      };
+    } else {
+      axiosDetails = {
+        baseURL: `${this.BASE_URL}/api/v4/f/${envId}`,
+        headers: {
+          "User-Agent": `silverfin-cli/${pkg.version}`,
+          Authorization: `Bearer ${firmTokens.accessToken}`,
```
agon-qurdina, line 62: Why not send accessToken as a `param` for both staging and production? It'd make it more obvious that the only difference between staging and production is really basic auth. Otherwise the regular API authentication works exactly the same.

agon-qurdina (approved):
Left a few comments but looks good in general. 

:100: for adding tests to it.

One thing I'd like to see is a bit more info added to Readme on how to use staging and a few words on how this solves the token conflicts (basic auth vs regular auth), considering this is both built and reviewed by team-recon but will be used by BSO.

lib/api/axiosFactory.js
```diff
@@ -0,0 +1,274 @@
...
+      async (error) => {
+        const originalConfig = error.config;
+        if (error.response) {
+          if (error.response.status === 401 && !originalConfig._retry) {
+            // Set _retry to true after trying the first refresh request on 401 status to avoid an infinite loop
+            originalConfig._retry = true;
+
+            try {
+              // Get a refreshed set of tokens
+              const firmId = originalConfig.baseURL.split("/").pop();
```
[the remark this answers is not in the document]
AgustinSilverfin: Good point. Yes, there should be. It was already implemented like this but I'll update it now :+1:

lib/api/axiosFactory.js
agon-qurdina: > Isn't headers preferred over param?

Haven't read anything suggesting that. Do you have any reference?

bin/cli.js
```diff
@@ -760,6 +760,15 @@ program
       "Get a new partner api key using the stored api key"
     )
   )
+  .addOption(
+    new Option(
+      "--set-host [host]",
+      "Set a custom host for the Silverfin API (e.g. https://live.getsilverfin.com)"
```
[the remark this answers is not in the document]
AgustinSilverfin: They usually use 2 already

lib/api/axiosFactory.js
agon-qurdina: I'd guess it's more of a convention than any real benefit, but let's leave it as is then.

lib/api/axiosFactory.js
AgustinSilverfin: Updated

lib/api/axiosFactory.js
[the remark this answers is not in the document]
AgustinSilverfin: Updated

PR #149, opened 2024-11-05
ReconciliationText#read - only write config when  it was changed

AgustinSilverfin:
## Description

When we use `ReconciliationText.read` we perform some checks on the config.json and write it back.
Add an extra check to write only if config has change, to avoid unnecessary write operations in the file system

For reference: https://silverfin.slack.com/archives/C07F1CEQP0S/p1730322904162779

## Type of change

- [X] Bug fix
- [] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/templates/reconciliationText.js
+++ b/lib/templates/reconciliationText.js
@@ -92,8 +92,10 @@ class ReconciliationText {
   static read(handle) {
     if (!templateUtils.checkValidName(handle, this.TEMPLATE_TYPE)) return false;
     fsUtils.createTemplateFolders(this.TEMPLATE_TYPE, handle, true);
+
     // Config.json
     let templateConfig = fsUtils.readConfig(this.TEMPLATE_TYPE, handle);
+    let originalConfig = { ...templateConfig };
     templateConfig = this.#checkHandleAndNameInConfig(templateConfig, handle);
     let template = this.#filterConfigItems(templateConfig);
     template = this.#checkReconciliationType(template);
@@ -104,7 +106,8 @@ class ReconciliationText {
     template.text = this.#readMainLiquid(handle, templateConfig);
     template.text_parts = this.#readPartsLiquid(handle, templateConfig);
 
-    fsUtils.writeConfig(this.TEMPLATE_TYPE, handle, templateConfig);
+    this.#updateConfigIfChanged(originalConfig, templateConfig, handle);
+
     return template;
   }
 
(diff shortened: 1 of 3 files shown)


lib/templates/reconciliationText.js
```diff
@@ -104,7 +106,8 @@ class ReconciliationText {
     template.text = this.#readMainLiquid(handle, templateConfig);
     template.text_parts = this.#readPartsLiquid(handle, templateConfig);
 
-    fsUtils.writeConfig(this.TEMPLATE_TYPE, handle, templateConfig);
+    this.#updateConfigIfChanged(originalConfig, templateConfig, handle);
```
Toby-Masters-SF, line 109: [COMMENT] New method

lib/templates/reconciliationText.js
```diff
@@ -92,8 +92,10 @@ class ReconciliationText {
   static read(handle) {
     if (!templateUtils.checkValidName(handle, this.TEMPLATE_TYPE)) return false;
     fsUtils.createTemplateFolders(this.TEMPLATE_TYPE, handle, true);
+
     // Config.json
     let templateConfig = fsUtils.readConfig(this.TEMPLATE_TYPE, handle);
+    let originalConfig = { ...templateConfig };
```
Toby-Masters-SF, line 98: [COMMENT] Use of Spread syntax, expanding an array, in this case the config

lib/templates/reconciliationText.js
```diff
@@ -92,8 +92,10 @@ class ReconciliationText {
...
     templateConfig = this.#checkHandleAndNameInConfig(templateConfig, handle);
```
Toby-Masters-SF, line 99: [SUGGESTION] Should orginalConfig be initialised _after_ `#checkHandleAndNameInConfig` potentially populates the fields `handle`, `name_en`, and `name_nl`?

lib/templates/reconciliationText.js
```diff
@@ -153,6 +156,12 @@ class ReconciliationText {
     }, {});
   }
 
+  static #updateConfigIfChanged(originalConfig, currentConfig, handle) {
```
Toby-Masters-SF, line 159: [SUGGESTION] Should this item be moved to `fsUtils` as may also wish to use for other template types?

Toby-Masters-SF (approved):
Hi Rufo, 
Approved with a couple of suggestions.
Kind regards,
Toby

lib/templates/reconciliationText.js
AgustinSilverfin: You are right that `#checkHandleAndNameInConfig` can modify the config, and that' what we want to check, so originalConfig should be defined before.

We define originalConfig as soon as templateConfig is loaded, kept as a copy of how it is currently stored. Later on we check it how it is at the end to check if it has changed or not, and only if changed we write again

lib/templates/reconciliationText.js
AgustinSilverfin: Good remark. I though about it, and currently no other template writes the config when calling `read()`, only ReconciliationText. 

I would say that makes more sense actually, we shouldn't write a new config when just reading. In ReconciliationText we are performing some adjustments. We even consider getting rid of those write operations and stick to reading only. But from now on we will only write if changed (once prrobably?)

PR #151, opened 2024-11-20
Not permitted to update new template

### LINKED ISSUE #121: Not permitted to update new template
**Describe the bug**
After creating a new reconciliation through the CLI, when I try to update that reconciliation I get an error.

 Response Error (422): {"downloadable_as_docx":"found unpermitted parameter: downloadable_as_docx for non externally managed template"} 
 You don't have the rights to update the previous parameters

I authorized the firm again but the issue persists.

**To Reproduce**
1. Create new recon. with the CLI
2. Update the main.liquid with some code
3. Run the update command through the CLI or try using development mode

**Expected behavior**
Recon. is updated in the firm

 - OS: macOS Sonoma 14.4
 - CLI version 1.26.4

**Additional context**

michieldegezelle:
## Description

Fix was made in line with what you proposed here: https://gitlab.silverfin.com/development/silverfin/-/issues/23082.

We will basically loop through all config items (with the reduce function), if the config item is present in the list of externally managed items (which is a new array) AND the config item externally_managed is put to false -> then the downloadable_as_docx will not be added to the final list of config items. 

This was verified within the be_market repo already, and seemed to work fine. downloadable_as_docx can be true/false, it won't be updated on the platform as long as externally_managed is put to false. 

Updating the template via the Silverfin extension doesn't work at the moment, but I assume this is because we still refer to the main version of the silverfin-cli code (thus excluding this check). Could you confirm this, and is there any way in which we can check this anyway? 

Fixes #121 

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/templates/reconciliationText.js
+++ b/lib/templates/reconciliationText.js
@@ -27,6 +27,9 @@ class ReconciliationText {
     "can_be_reconciled_without_data",
     "only_reconciled_with_data",
   ];
+  static EXTERNALLY_MANAGED_ITEMS = [
+    "downloadable_as_docx"
+  ];
   static TEMPLATE_TYPE = "reconciliationText";
   static TEMPLATE_FOLDER = fsUtils.FOLDERS[this.TEMPLATE_TYPE];
   constructor() {}
@@ -149,6 +152,12 @@ class ReconciliationText {
 
   static #filterConfigItems(templateConfig) {
     return this.CONFIG_ITEMS.reduce((acc, attribute) => {
+      // Skip externally managed items if externally_managed is false
+      if (this.EXTERNALLY_MANAGED_ITEMS.includes(attribute) &&
+        templateConfig.externally_managed === false) {
+        return acc;
+      }
+
       if (templateConfig.hasOwnProperty(attribute)) {
         acc[attribute] = templateConfig[attribute];
       }
--- a/lib/templates/reconciliationText.js
+++ b/lib/templates/reconciliationText.js
@@ -151,10 +151,14 @@ class ReconciliationText {
   }
 
   static #filterConfigItems(templateConfig) {
-    return this.CONFIG_ITEMS.reduce((acc, attribute) => {
+
+    const skippedItems = [];
+
+    const filteredConfig = this.CONFIG_ITEMS.reduce((acc, attribute) => {
       // Skip externally managed items if externally_managed is false
       if (this.EXTERNALLY_MANAGED_ITEMS.includes(attribute) &&
         templateConfig.externally_managed === false) {
+        skippedItems.push(attribute);
         return acc;
       }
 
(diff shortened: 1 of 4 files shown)


AgustinSilverfin (commented):
Update looks good and seems to be working. I have two comments to consider:

- Now the item is skipped but the user may not notice it. Can we show a message (using consola.warn maybe) indicating which items were skipped? Do you think that would be valuable? It could look like this:
 
image

- Let's consider and add some test for this update as well (considering we already have tests for `ReconciliationText#read`, we can look into expanding those to test this)

> Updating the template via the Silverfin extension doesn't work at the moment, but I assume this is because we still refer to the main version of the silverfin-cli code (thus excluding this check). Could you confirm this, and is there any way in which we can check this anyway?

Good point, the extension uses it's own version of the cli as a dependency. Releasing a new version of the cli won't impact automatically the extension. We need to go to the extension repository and bump the cli version when it's needed, and deploy it to vscode marketplace. We can look into this once this PR is released, and we go through the entire release process of the extension

michieldegezelle:
@AgustinSilverfin Consola warning + test was added. Could you please do another review?

lib/templates/reconciliationText.js
```diff
@@ -155,6 +155,9 @@ class ReconciliationText {
       // Skip externally managed items if externally_managed is false
       if (this.EXTERNALLY_MANAGED_ITEMS.includes(attribute) &&
         templateConfig.externally_managed === false) {
+        consola.warn(
+          `The following attributes were skipped because they can only be updated when the template is externally managed: ${this.EXTERNALLY_MANAGED_ITEMS.join(", ")}.`
+        );
```
AgustinSilverfin, line 160: This is good, but if we include it here (in the iteration itself) it will be printed each time there is a skipped item.
Now it's only once because there is only one item `downloadable_as_docx`. But if we ever add more items the message will be repeated as many times as items are skipped

tests/lib/templates/reconciliationTexts.test.js
```diff
@@ -452,5 +452,22 @@ describe("ReconciliationText", () => {
 
       expect(result.text_parts).toEqual([{ name: "empty_part", content: "" }]);
     });
+
+    it("should exclude downloadable_as_docx and show warning when externally_managed is false", () => {
```
AgustinSilverfin, line 456: This one looks good :+1: 
Maybe we can add a similar one testing that it's included when externally_managed is true?

lib/templates/reconciliationText.js
michieldegezelle: I have adjusted the logic so that the warning is only shown outside of the reduce.

tests/lib/templates/reconciliationTexts.test.js
michieldegezelle: Positive scenario has been added as well

PR #152, opened 2024-11-22
Fix add-shared-part --all command

AgustinSilverfin:
## Description

This command was 'broken' when Partners API was implemented. 
Force flag was still accepted as argument but never used Await clause and delay were removed
Here we recover the implementation that we have before Partners API, adding the corresponding options to select between firm and partner ids when corresponding
Based on commit 6c4cd17

Fixes 

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -949,9 +949,14 @@ async function addSharedPart(
  * @param {boolean} force - If true, it will add the shared part to all templates, even if it's already present
  */
 async function addAllSharedParts(type, envId, force = false) {
+  const envConfigKey = type === "partner" ? "partner_id" : "id";
   const sharedPartsArray = fsUtils.getAllTemplatesOfAType("sharedPart");
-  for (let sharedPartName of sharedPartsArray) {
-    let sharedPartConfig = fsUtils.readConfig("sharedPart", sharedPartName);
+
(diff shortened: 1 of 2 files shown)


delphine-demeulenaere (commented):
Hi Rufo,

I noticed that you didn't include a commit to update the dependencies and adjust the version number in the `package.json` file. Was there a specific reason for leaving this out? Because I thought this was necessary for testing?

AgustinSilverfin:
> Hi Rufo,
> 
> I noticed that you didn't include a commit to update the dependencies and adjust the version number in the `package.json` file. Was there a specific reason for leaving this out? Because I thought this was necessary for testing?

@delphine-demeulenaere no reason, I just forgot I guess :slightly_smiling_face: 

Also version should be always checked right before merging because there could be other PRs merged in the mean time affecting the latest version

delphine-demeulenaere (approved):
Everything appears to be functioning correctly during testing! 

**Note**: @AgustinSilverfin I approved the changes, but I’m not entirely sure if they should undergo another review by team-recon for confirmation.

PR #153, opened 2024-12-04
Support export files on partners

michieldegezelle:
## Description

Update to make sure that export files can be handled (imported & updated) on partners as well. 
Note that the create commands were not yet added, since we have a separate ticket for that to cover all template types.  

Fixes # [[Liquid toolchain] - CLI - Support export files on Partners](https://gitlab.silverfin.com/development/silverfin/-/issues/23085)

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -105,7 +105,7 @@ program
       );
     } else if (options.all) {
       toolkit.publishAllReconciliations(
-        "firm",
+        settings.type,
         settings.envId,
         options.message
       );
@@ -210,7 +210,7 @@ program
       );
     } else if (options.all) {
       toolkit.publishAllExportFiles(
-        "firm",
+        settings.type,
         settings.envId,
         options.message
       );
@@ -318,7 +318,7 @@ program
       );
     } else if (options.all) {
       toolkit.publishAllAccountTemplates(
-        "firm",
+        settings.type,
         settings.envId,
         options.message
       );
@@ -425,7 +425,11 @@ program
         options.message
       );
     } else if (options.all) {
-      toolkit.publishAllSharedParts("firm", settings.envId, options.message);
+      toolkit.publishAllSharedParts(
+        settings.type, 
+        settings.envId, 
+        options.message
+      );
     }
   });
 
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -145,6 +145,7 @@ program
   .command("import-export-file")
   .description("Import export file templates")
   .option("-f, --firm <firm-id>", "Specify the firm to be used", firmIdDefault)
+  .option("-p, --partner <partner-id>", "Specify the partner to be used")
   .option("-n, --name <name>", "Import a specific export file by name")
   .option("-i, --id <id>", "Import a specific export file by id")
   .option("-a, --all", "Import all existing export files")
@@ -158,7 +159,7 @@ program
       ["name", "id", "all", "existing"],
       options,
       firmIdDefault,
-      false
+      true // Partner supported
     );
 
     if (options.name) {
@@ -168,11 +169,15 @@ program
         options.name
       );
     } else if (options.id) {
-      toolkit.fetchExportFileById(settings.type, settings.envId, options.id);
+      toolkit.fetchExportFileById(
+        settings.type, 
+        settings.envId, 
+        options.id
+      );
     } else if (options.all) {
       toolkit.fetchAllExportFiles(settings.type, settings.envId);
-    } else if (options.all) {
-      toolkit.fetchExistingExportFiles(options.firm);
+    } else if (options.existing) {
+      toolkit.fetchExistingExportFiles(settings.type, settings.envId);
     }
   });
 
@@ -181,6 +186,7 @@ program
   .command("update-export-file")
   .description("Update an existing export file template")
   .option("-f, --firm <firm-id>", "Specify the firm to be used", firmIdDefault)
+  .option("-p, --partner <partner-id>", "Specify the partner to be used")
   .option("-n, --name <name>", "Specify the export file to be used (mandatory)")
   .option("-a, --all", "Update all export files")
   .option(
@@ -194,8 +200,8 @@ program
       ["name", "all"],
       options,
       firmIdDefault,
-      false,
-      true // Message required (added for later)
+      true, // Partner supported
+      true // Message required
     );
 
     if (options.name) {
@@ -207,7 +213,7 @@ program
       );
     } else if (options.all) {
       toolkit.publishAllExportFiles(
-        settings.type,
+        "firm",
         settings.envId,
         options.message
       );
@@ -237,7 +243,7 @@ program
     if (options.name) {
       toolkit.newExportFile("firm", options.firm, options.name);
     } else if (options.all) {
-      toolkit.newAllExportFiles("firm", options.firm, options.name);
+      toolkit.newAllExportFiles("firm", options.firm);
     }
   });
 
@@ -488,12 +494,11 @@ program
   )
   .option("--yes", "Skip the prompt confirmation (optional)")
   .action((options) => {
-    const partnerSupported = options.exportFile ? false : true;
     const settings = runCommandChecks(
       ["sharedPart", "all"],
       options,
       firmIdDefault,
-      partnerSupported
+      true // Partner supported
     );
 
     if (options.sharedPart) {
@@ -560,12 +565,11 @@ program
   )
   .option("--yes", "Skip the prompt confirmation (optional)")
   .action((options) => {
-    const partnerSupported = options.exportFile ? false : true;
     const settings = runCommandChecks(
       ["handle", "exportFile", "accountTemplate"],
       options,
       firmIdDefault,
-      partnerSupported
+      true // Partner supported
     );
 
     if (options.handle) {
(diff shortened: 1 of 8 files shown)


bin/cli.js
```diff
@@ -158,7 +159,6 @@ program
       ["name", "id", "all", "existing"],
       options,
       firmIdDefault,
-      false
```
delphine-demeulenaere, line 161: add true

bin/cli.js
```diff
@@ -207,7 +212,7 @@ program
       );
     } else if (options.all) {
       toolkit.publishAllExportFiles(
-        settings.type,
+        "partner",
```
delphine-demeulenaere, line 215: should be changed to "firm" instead of "partner"

bin/cli.js
```diff
@@ -894,7 +898,6 @@ program
       ["name", "all"],
       options,
       firmIdDefault,
-      false
```
delphine-demeulenaere, line 897: add true

index.js
```diff
@@ -280,25 +292,44 @@ async function fetchExportFileById(type, envId, id) {
...
+  }
+
   templates.forEach(async (name) => {
     const templateConfig = fsUtils.readConfig("exportFile", name);
-    if (!templateConfig || !templateConfig.id[firmId]) return;
-    await fetchExportFileById(firmId, templateConfig.id[firmId]);
+    let templateId = 
+      type == "firm"
+        ? templateConfig?.id?.[envId]
+        : templateConfig?.partner_id?.[envId]
```
AgustinSilverfin, line 325: I know this is how it's implemented already, but I would love to see this moved (from everywhere) into a helper function

```js
function getTemplateId(type, envId, templateConfig) {
  switch (type) {
    case "firm":
      return templateConfig?.id?.[envId];
      break;
    case "partner":
      return templateConfig?.partner_id?.[envId];
      break;
    default:
      throw "Invalid template type: ${type}";
      break;
}
```
And we store in one of the utils files, where we have other config helpers.
What do you think? (Also, not completly sure it works exactly as it is, I just wrote it here while doing the review :slightly_smiling_face: )

we do later:
```js
let templateId = fsUtils.getTemplateId(type, envId, templateConfig);
```

index.js
```diff
@@ -461,7 +493,11 @@ async function fetchAllAccountTemplates(type, envId, page = 1) {
 
 async function fetchExistingAccountTemplates(type, envId) {
   const templates = fsUtils.getAllTemplatesOfAType("accountTemplate");
-  if (!templates) return;
+  if (!templates) {
+    consola.warn("No account templates found");
```
AgustinSilverfin, line 497: :100:

index.js
```diff
@@ -782,8 +818,8 @@ async function newAllSharedParts(type, firmId) {
 }
 
 /** This function adds a shared part to a template. It will make a POST request to the API. If the ID of one of the templates is missing, it will try to fetch it first by making a GET request. In case of success, it will store the details in the corresponding config files.
- *
- * @param {Number} firmId
+ * @param {String} type - Options: `firm` or `partner`
```
AgustinSilverfin, line 821: :100:

index.js
```diff
@@ -797,14 +833,6 @@ async function addSharedPart(
...
 ) {
   try {
-    // Add a check for export files that are not supported
-    if (type == "partner" && templateType == "exportFile") {
-      consola.warn(
-        "Adding shared parts to export files on partner is not supported. Skipping."
-      );
-      return false;
-    }
-
```
AgustinSilverfin, line 807: :tada:

lib/templates/exportFile.js
```diff
@@ -17,7 +17,8 @@ class ExportFile {
 
   /**
    * Process the response provided by the Silverfin API and store every detail in its corresponding file (liquid files, config file, etc)
-   * @param {number} firmId
+   * @param {string} type firm or partner
+   * @param {number} envId  The id of the firm or partner environment where the template is going to be imported from
    * @param {object} template
    */
   static save(type, envId, template) {
```
AgustinSilverfin, line 24: This is probably why you removed the `await` clauses.

This may be something to monitor an align, now `ReconciliationText` and `SharedParts` have `async save` while `ExportFile` and `AccountTemplate` doesn't :thinking:

lib/templates/exportFile.js
```diff
@@ -46,9 +47,6 @@ class ExportFile {
         ...existingConfig?.partner_id,
         ...addNewId(type, "partner", envId, template),
       },
-      partner_id: {
-        ...existingConfig?.partner_id,
-      },
```
AgustinSilverfin, line 51: I guess we are still finding things from the Partners API update :+1: :joy:

index.js
```diff
@@ -280,25 +292,44 @@ async function fetchExportFileById(type, envId, id) {
...
     if (page == 1) {
-      consola.error(`No export files found in firm ${firmId}`);
+      consola.error("No export files found in firm");
     }
     return;
   }
+
   templates.forEach(async (template) => {
-    fetchExportFileById(type, envId, template.id);
+    const saved = ExportFile.save(type, envId, template.id);
```
AgustinSilverfin, line 304:
```suggestion
    const saved = ExportFile.save(type, envId, template);
```
image

bin/cli.js
```diff
@@ -207,7 +213,7 @@ program
       );
     } else if (options.all) {
       toolkit.publishAllExportFiles(
-        settings.type,
+        "firm",
```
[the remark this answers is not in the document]
AgustinSilverfin: Also the error when running `silverfin update-<any-template-type> --all --partner <id> --message "Update all` is not clear. 

It is not saying it is not supported, it just says missing IDs.

I would vote for change it now, replace "firm" with `type` and support it for every `update-` command. What do you think?

lib/cli/utils.js
```diff
@@ -142,13 +142,7 @@ function runCommandChecks(
   messageRequired = false
 ) {
   if (options.partner) {
-    // To delete once export files are supported
-    if (!partnerSupported) {
-      consola.error(
-        `The option "--partner" is not supported when using "--export-file"`
-      );
-      process.exit(1);
-    } else if (messageRequired && !options.message) {
```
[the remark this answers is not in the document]
michieldegezelle: I removed the parameter from the function and all the references (true) to it in the underlying commands.

bin/cli.js
```diff
@@ -168,11 +169,15 @@ program
...
-      toolkit.fetchExportFileById(settings.type, settings.envId, options.id);
+      toolkit.fetchExportFileById(
+        settings.type, 
+        settings.envId, 
+        options.id
+      );
     } else if (options.all) {
       toolkit.fetchAllExportFiles(settings.type, settings.envId);
-    } else if (options.all) {
-      toolkit.fetchExistingExportFiles(options.firm);
```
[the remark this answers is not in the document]
michieldegezelle: We had two instances of the options.all (and no .existing even though the function was the one for existing export files) 😄

index.js
```diff
@@ -280,25 +292,44 @@ async function fetchExportFileById(type, envId, id) {
 
 async function fetchAllExportFiles(type, envId, page = 1) {
   const templates = await SF.readExportFiles(type, envId, page);
+
   if (templates.length == 0) {
     if (page == 1) {
-      consola.error(`No export files found in firm ${firmId}`);
+      consola.error("No export files found in firm");
```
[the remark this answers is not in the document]
michieldegezelle: Added it to all possible command to give as much information to the user as possible

index.js
```diff
@@ -280,25 +292,44 @@ async function fetchExportFileById(type, envId, id) {
...
   });
   fetchAllExportFiles(type, envId, page + 1);
 }
 
-async function fetchExistingExportFiles(firmId) {
+async function fetchExistingExportFiles(type, envId) {
   const templates = fsUtils.getAllTemplatesOfAType("exportFile");
-  if (!templates) return;
+  if (!templates) {
+    consola.warn("No export files found");
```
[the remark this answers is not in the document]
michieldegezelle: Added it to all possible command to give as much information to the user as possible

index.js
michieldegezelle: Removed the `.id`

index.js
michieldegezelle: Implemented it as requested

AgustinSilverfin (changes requested):
I left only one comment regarding partners and update/create all commands to decide what we do.

Besides that, everything looks good for me. Which takes me to the next point. Did someone else from the team did a functional testing? if not, we should do that

index.js
AgustinSilverfin, line 1: Sorry @BenjaminLangenakenSF , but I would prefer to have all this conditions on `/bin/cli.js` rather than `index.js`
I see this contrain related to the CLI commands & flags, so we add it there, while keeping the functionality 'clean'

If we ever support them, again we update CLI commands & flags.

What do you think?

michieldegezelle: Hi @AgustinSilverfin 

How would that look like? Something like this? 
image

So adding a condition to check on the combination of those options and then a process.exit?

AgustinSilverfin: Yes.
Or even cleaner (to avoid repeating always the same and adding multiple lines) we create a helper method like     `cliUtils.checkUniqueOption(["handle", "all"], options)` or `cliUtils.checkDefaultFirm(options.firm, firmIdDefault)` where we pass options

E.g. `cliUtils.checkPartnerSupport(options)` I believe the condition is options.all and options.type == partner, right?

BenjaminLangenakenSF: Thanks for the explanation Rufo, added a fixup commit 😉
Looks ok?

AgustinSilverfin (approved):
LGTM!

PR #154, opened 2024-12-05
Fix: authorization when there are no stored tokens

AgustinSilverfin:
## Description

All AxiosInstances will check that there stored access and refresh tokens for the firm, if not it will raise an error to authorize. For that reason, when we are actually authorizating, we don't have to use those instance. If not, the first time you authorize a firm an error will be raised before you can actually authorize it.

## How to test it

- Choose a demo firm and log in into Silverfin in the browser
- Make sure you don't have any old credentials for it (remove it from the config file if needed)
- Run a new `silverfin authorize`

## Type of change

- [X] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/utils/apiUtils.js
+++ b/lib/utils/apiUtils.js
@@ -26,7 +26,7 @@ function checkRequiredEnvVariables() {
 }
 
 // Get Tokens for the first time with an authorization code
-async function getAccessToken(axiosInstance, firmId, authCode) {
+async function getAccessToken(firmId, authCode) {
   try {
     const redirectUri = "urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob";
     const grantType = "authorization_code";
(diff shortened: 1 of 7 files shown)


lib/utils/apiUtils.js
```diff
@@ -26,15 +26,15 @@ function checkRequiredEnvVariables() {
...
+async function getAccessToken(firmId, authCode) {
   try {
     const redirectUri = "urn%3Aietf%3Awg%3Aoauth%3A2.0%3Aoob";
     const grantType = "authorization_code";
     let requestDetails = {
       method: "POST",
       url: `${BASE_URL}/f/${firmId}/oauth/token?client_id=${process.env.SF_API_CLIENT_ID}&client_secret=${process.env.SF_API_SECRET}&redirect_uri=${redirectUri}&grant_type=${grantType}&code=${authCode}`,
     };
-    const response = await axiosInstance(requestDetails);
+    const response = await axios(requestDetails);
```
AgustinSilverfin: @delphine-demeulenaere This line was the issue.

When I prepared the staging update, I replaced this line as well to use the new `axiosInstance`, but it wasn't using it's equivalent before.

So I "revert" it to use it `axios` module directly, without the instances we create for other API calls. That is, with no interceptors in place to check existing tokens and refresh tokens if needed.

All other changes are unrelated, to improve redability, isolate logic into it's own class and write test for the authorization process :slightly_smiling_face:

delphine-demeulenaere: Oh, I see! That makes sense. I don’t think I would have discovered this issue on my own though 😅 I tested it and it seems to be working as expected! Thank you Rufo for addressing this so quickly

delphine-demeulenaere (approved):

PR #156, opened 2024-12-06
Refactor refresh credentials methods

AgustinSilverfin:
## Description

Refactoring with the goal of improving the code quality, by grouping related functionality, in this case refresh credentials methods within the previously created `SilverfinAuthorizer` class, and adding test coverage for the mentioned functionality.

The `SilverfinAuthorizer` class should handle all Authorizations, which includes renewing (refreshing) expired credentials manually.
Also, all axios instances should be generated by `AxiosFactory`, which will handle to add headers if needed (basic auth for stagings, for example) and request interceptors to refresh token on the fly if they have expired.

Considering that all axios instances will be generated by AxiosFactory, there is a new method to create instances dedicated to the authorization process, which does not validate the existence of previous tokens and it does not add any interceptors

## Test

Make sure that the following commands works both in Production and Staging

```
silverfin authorize
silverfin authorize-partner
silverfin config --refresh-token
silverfin config --refresh-partner-token
```

## Type of change

- [X] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -756,13 +756,13 @@ program
   )
   .addOption(
     new Option(
-      "--refresh-partner-token [partner_id]",
+      "--refresh-partner-token <partnerId>",
       "Get a new partner api key using the stored api key"
     )
   )
   .addOption(
     new Option(
-      "--set-host [host]",
+      "--set-host <host>",
       "Set a custom host for the Silverfin API (e.g. https://live.getsilverfin.com)"
     )
   )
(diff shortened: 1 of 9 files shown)


michieldegezelle (dismissed):
All commands seem to work fine! 
Just 1 potential improvement: in case I try to import something from a firm that is not yet authorized, I get a clear warning that I should authorize first. If I try to do the same for partners, the terminal just keeps loading without showing any warning. I have to exit it with CTRL + C, which is not the most user-friendly way. 

Could you look into showing a message in this case as well?

AgustinSilverfin:
> All commands seem to work fine! Just 1 potential improvement: in case I try to import something from a firm that is not yet authorized, I get a clear warning that I should authorize first. If I try to do the same for partners, the terminal just keeps loading without showing any warning. I have to exit it with CTRL + C, which is not the most user-friendly way.
> 
> Could you look into showing a message in this case as well?

Can you test this again when you have a moment? @michieldegezelle

delphine-demeulenaere (approved):
I tested the commands again, and everything appears to be working correctly. I now also receive the error message!

Issue #145, opened 2024-10-24
New features to fully automate deployment [MYOB]


RomySilverfin:
**Is your feature request related to a problem? Please describe.**
To be able to automate the deployment of updates or new development, some steps in the chain are currently missing.
1. The option to create templates and shared parts via de CLI in a partner environment
2. The option to update or create **all** templates and shared parts that are part of the pull request in the partner environment

**Describe the solution you'd like**
1. create-reconciliation, create-shared-part, create-account-template etc. should work for partner enviroments too
2. adding --all to create & update commands when pushing to partners

**Describe alternatives you've considered**
1. Manual workaround: use special action to create new template on partners
2. Write a script that filters out the handles/identifiers of the templates in the PR and create separate commands for each of them.

**Additional context**
MYOB aims to fully automate as much as they can when it comes to releasing updates.
They want to avoids manual errors. They have a lot of different software systems they need to keep up-to-date and want to streamline as much as possible.
The steps mentioned above are a considerable blocker.

BenjaminLangenakenSF:
Scope update — half of this has landed, the other half is currently refused by design.

**Part 1 (create on partner) — done.** Every `create-*` command now accepts `-p, --partner`: `create-reconciliation`, `create-export-file`, `create-account-template`, `create-shared-part`.

**Part 2 (`--all` on partner) — blocked deliberately.** [`lib/cli/utils.js#L191-L196`](https://github.com/silverfin/silverfin-cli/blob/b1991fcf931e8e2fb3b06a4e26744507d0a5f756/lib/cli/utils.js#L191-L196):

```js
if (options.partner && options.all) {
  consola.error("Not possible to update all templates at once in a partner environment.");
  process.exit(1);
}
```

`create-all-templates` / `update-all-templates` are also firm-only — no `--partner` option at all.

So what's left isn't implementation work, it's a decision: should bulk pushes to a partner be allowed, and under what guard? Worth answering before this is picked up.

PR #158, opened 2024-12-18
Support create commands

michieldegezelle:
## Description

Create commands were updated to work for partners as well (reconciliation texts, account templates and shared parts). 

Fixes # https://github.com/silverfin/silverfin-cli/issues/145

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -259,6 +259,7 @@ program
     "Try to create all export files stored in the repository"
   )
   .action((options) => {
+    cliUtils.checkPartnerSupport(options);
     const settings = runCommandChecks(
       ["name", "all"],
       options,
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -135,6 +135,7 @@ program
     "Try to create all the reconciliation texts stored in the repository"
   )
   .action((options) => {
+    cliUtils.checkPartnerSupport(options);
     const settings = runCommandChecks(
       ["handle", "all"],
       options,
@@ -379,6 +380,7 @@ program
     "Try to create all account templates stored in the repository"
   )
   .action((options) => {
+    cliUtils.checkPartnerSupport(options);
     const settings = runCommandChecks(
       ["name", "all"],
       options,
@@ -502,6 +504,7 @@ program
     "Try to create all the shared parts stored in the repository"
   )
   .action((options) => {
+    cliUtils.checkPartnerSupport(options);
     const settings = runCommandChecks(
       ["sharedPart", "all"],
       options,
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -470,10 +470,14 @@ program
 program
   .command("create-shared-part")
   .description("Create a new shared part")
-  .requiredOption(
+  .option(
     "-f, --firm <firm-id>",
     "Specify the firm to be used",
     firmIdDefault
+  )
+  .option(
+    "-p, --partner <partner-id>",
+    "Specify the partner to be used"
   )
   .option(
     "-s, --shared-part <name>",
@@ -484,12 +488,20 @@ program
     "Try to create all the shared parts stored in the repository"
   )
   .action((options) => {
-    cliUtils.checkUniqueOption(["sharedPart", "all"], options);
-    cliUtils.checkDefaultFirm(options.firm, firmIdDefault);
+    const settings = runCommandChecks(
+      ["sharedPart", "all"],
+      options,
+      firmIdDefault)
+
     if (options.sharedPart) {
-      toolkit.newSharedPart("firm", options.firm, options.sharedPart);
+      toolkit.newSharedPart(
+        settings.type, 
+        settings.envId, 
+        options.sharedPart);
     } else if (options.all) {
-      toolkit.newAllSharedParts("firm", options.firm);
+      toolkit.newAllSharedParts(
+        settings.type, 
+        options.envId);
     }
   });
 
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -347,10 +347,14 @@ program
 program
   .command("create-account-template")
   .description("Create a new account template")
-  .requiredOption(
+  .option(
     "-f, --firm <firm-id>",
     "Specify the firm to be used",
     firmIdDefault
+  )
+  .option(
+    "-p, --partner <partner-id>",
+    "Specify the partner to be used"
   )
   .option(
     "-n, --name <name>",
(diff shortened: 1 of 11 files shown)


index.js
```diff
@@ -205,42 +205,48 @@ async function publishAllReconciliations(
...
   }
 }
 
-async function newAllReconciliations(type, firmId) {
+async function newAllReconciliations(type, envId) {
+
+  if (type == "partner") {
+    consola.error(`Not possible to create all reconciliations at once in a partner environment`)
+    return;
+  }
```
AgustinSilverfin, line 245: Let's use the same approach as in #153.

` cliUtils.checkPartnerSupport(options);` in each command in `bin/cli.js`

index.js
```diff
@@ -371,34 +377,40 @@ async function publishAllExportFiles(
...
   }
 }
 
-async function newAllExportFiles(type, firmId) {
+async function newAllExportFiles(type, envId) {
+
+  if (type == "partner") {
+    consola.error(`Not possible to create all export files at once in a partner environment`)
+    return;
+  }
```
AgustinSilverfin, line 409: Move to `bin/cli.js`

index.js
```diff
@@ -577,23 +590,29 @@ async function newAccountTemplate(type, firmId, name) {
...
   }
 }
 
-async function newAllAccountTemplates(type, firmId) {
+async function newAllAccountTemplates(type, envId) {
+
+  if (type == "partner") {
+    consola.error(`Not possible to create all account templates at once in a partner environment`)
+    return;
+  }
```
AgustinSilverfin, line 611: Move to `bin/cli.js`

index.js
```diff
@@ -746,38 +765,45 @@ async function publishAllSharedParts(
...
   }
 }
 
-async function newAllSharedParts(type, firmId) {
+async function newAllSharedParts(type, envId) {
+
+  if (type == "partner") {
+    consola.error(`Not possible to create all shared parts at once in a partner environment`)
+    return;
+  }
```
AgustinSilverfin, line 802: Move to `bin/cli.js`

lib/templates/sharedPart.js
```diff
@@ -69,9 +69,14 @@ class SharedPart {
...
   /** Update template's id for the corresponding firm in template's config file */
-  static updateTemplateId(firmId, name, templateId) {
+  static updateTemplateId(type, envId, name, templateId) {
     let templateConfig = fsUtils.readConfig(this.TEMPLATE_TYPE, name);
-    templateConfig.id[firmId] = templateId;
+    if (type == "firm") {
+      templateConfig.id[envId] = templateId;
+    } else if (type == "partner") {
+      templateConfig.partner_id[envId] = templateId;
+    }
```
AgustinSilverfin, line 78: Considering that we are already creating a `getTemplateId` in `fsUtils` we may want to create a `setTemplateId` (To use in all template types), and also with the `default` statement that throw an "invalid template type error" (that error could be useful to debug if we ever make changes that breaks this `updateTemplateId` calling a wrong type for example :smile:

AgustinSilverfin (changes requested):

michieldegezelle:
@AgustinSilverfin Review comments have been implemented

lib/utils/fsUtils.js
```diff
@@ -454,13 +454,23 @@ function getTemplateId(type, envId, templateConfig) {
...
+function setTemplateId(type, envId, templateConfig, templateId) {
+  switch (type) {
+    case "firm":
+      templateConfig.id[envId] = templateId;
+      return templateConfig.id[envId];
+    case "partner":
+      templateConfig.partner_id[envId] = templateId;
+      return templateConfig.partner_id[envId];
+    default:
+      throw "Invalid template type: ${type}";
```
AgustinSilverfin, line 473: Why `break` statements are return? I believe they should be there. https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/switch

bin/cli.js
```diff
@@ -237,12 +257,22 @@ program
     "Try to create all export files stored in the repository"
   )
   .action((options) => {
-    cliUtils.checkUniqueOption(["name", "all"], options);
-    cliUtils.checkDefaultFirm(options.firm, firmIdDefault);
+    const settings = runCommandChecks(
+      ["name", "all"],
+      options,
+      firmIdDefault
+    );
```
AgustinSilverfin, line 264: By adding `runCommandChecks` to create commands, we are now always forced to confirm. should we add the `---yes` flag so we can skip confirmation?

AgustinSilverfin (commented):
@michieldegezelle besides those two comments, could you assign  this to someone else to have a look as well and to a functional testing?

bin/cli.js
michieldegezelle: What exactly did you have in mind here? Would this work? 
image

michieldegezelle:
@AgustinSilverfin Not 100% sure how to always add the 'yes' option, so proposed something (see image above). Once that comment is processed, I will assign it for testing.

bin/cli.js
AgustinSilverfin: Yes, that's what I had I mind BUT now I think it does not make sense :smile: 
We didn't have the `--yes` flag before for a reason. Create will never overwrite anything. If template already exists, it won't do nothing.
So I think that we need to skip the prompt confirmation completely, without any new flag.
This is one of the reason why I don't like this `runCommandChecks`, because it does many things and now we need another exception (in create commands, we don't need to check the yes option).

What do you think?

michieldegezelle:
@AgustinSilverfin Can be reviewed again

bin/cli.js
```diff
@@ -237,12 +259,24 @@ program
     "Try to create all export files stored in the repository"
   )
   .action((options) => {
-    cliUtils.checkUniqueOption(["name", "all"], options);
-    cliUtils.checkDefaultFirm(options.firm, firmIdDefault);
+    const settings = runCommandChecks(
```
AgustinSilverfin, line 262: Should we have checkPartnerSupport? It is in all other create commands

```suggestion
    cliUtils.checkPartnerSupport(options);
    const settings = runCommandChecks(
```

AgustinSilverfin (approved):
Looks good to me. Just look at the one template I left for partner support check.

I'll approve it but have some to do a functional test before merging

PR #165, opened 2025-02-06
Improve --help guidance for account templates and export files

BenjaminLangenakenSF:
## Description

Improved the --help guidance for account templates and export files to always use double quotes around the name.
e.g. `-n, --name "<name>"       Import a specific export file by name | Make sure to always enclose the name in double quotes`

## Type of change

- [ ] Bug fix
- [x] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -138,7 +138,7 @@ program
   .description("Import export file templates")
   .option("-f, --firm <firm-id>", "Specify the firm to be used", firmIdDefault)
   .option("-p, --partner <partner-id>", "Specify the partner to be used")
-  .option("-n, --name <name>", "Import a specific export file by name")
+  .option('-n, --name "<name>"', "Import a specific export file by name | Make sure to always enclose the name in double quotes")
   .option("-i, --id <id>", "Import a specific export file by id")
   .option("-a, --all", "Import all existing export files")
   .option("-e, --existing","Import all export files (already stored in the repository)")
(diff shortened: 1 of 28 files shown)


bin/cli.js
```diff
@@ -34,16 +34,10 @@ program
   .description("Import reconciliation templates")
   .option("-f, --firm <firm-id>", "Specify the firm to be used", firmIdDefault)
   .option("-p, --partner <partner-id>", "Specify the partner to be used")
-  .option(
-    "-h, --handle <handle>",
-    "Import a specific firm reconciliation by handle"
-  )
+  .option("-h, --handle <handle>","Import a specific firm reconciliation by handle")
```
AgustinSilverfin, line 37: This seems to be just a formatting difference, right? @michieldegezelle @BenjaminLangenakenSF 
Maybe we need to look into introduce some formatting rules so we don't have all this changes in files. I'll look into tihs

BenjaminLangenakenSF: Indeed made all the `.options` inline to make the code a bit more readable, but maybe personal preference 😅
Don't know if there are certain guidelines or coding practices to follow here?

AgustinSilverfin: Don't get me wrong, I also prefer it inline.
But as soon as I open this file in my editor and hit save... :smile: 
image
I think we can just add a rule in the repository that the formatter can pick it up so we avoid re-writting back

AgustinSilverfin: Prettier, indeed. I would suggest you to add it as well :smile: 
For example,  now your update is missing with spaces before descriptions.
It makes it easier to align this things across people.

package.json
AgustinSilverfin, line 1: If you look here we already have some basic prettier rules.
You can add `printWidth: 180` so it is inline with the way you formatted the descriptions.

Maybe we can do a follow up PR just to run the formatter to align this new rule to all files.

`npx prettier --write "**/*.js"` should do the trick

bin/cli.js
BenjaminLangenakenSF: Ok installed!

package.json
BenjaminLangenakenSF: Added two commits reflecting these changes 😉

AgustinSilverfin: And this is how we get rid of ~2K lines of code :joy: 
image

AgustinSilverfin (approved):

PR #166, opened 2025-02-06
Implement export file new name locales

michieldegezelle:
## Description

Used the specs here: https://gitlab.silverfin.com/development/silverfin/-/issues/24034

Couple of considerations:

- Do we want to remove the 'name' attribute from the config files as soon as someone does a 'import-export-file'? Now, we only add the 3 new locales. 

- When doing an update via 'update-export-file', some new logic was added to adjust the config file and add the missing name translations. This leads to 'new' changes after we do the update command (which can in some cases e.g. update to partners be the last action). As a result, we can end up with a situation where a fix was approved, merged with main, and when someone wants to push from main with the update command, we add the new locales. These new changes need to be put in a new PR to be reflected correctly in the repo's. Do you see any issues with this? This is a one-off of course. 

- Should we consider renaming the folder automatically to the nl naming as soon as we do any updates? 

- We have two main 'error paths':
In the update command, we still refer to the 'old' name, which is also still the folder name. This case is covered by a warning that the folder should be renamed

In the update command, we refer to the new nl name, but the folder name is still referring to the old name. We end up in the same situation as when we have a typo: Warning about ID that is missing. 

Given that this is a one-off, instead of having to write some additional logic here, could be maybe make the warning a little bit more descriptive?

So: ID is missing, aborted. Please check your command for typos and check if the folder name matches the name_nl (or something like this). 

- Do we still need the 'name' attribute in the config in the ExportFile class? 

## Type of change

- [ ] Bug fix
- [ ] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/index.js
+++ b/index.js
@@ -211,7 +211,7 @@ async function fetchExportFileByName(type, envId, name) {
 
     const saved = ExportFile.save(type, envId, template);
     if (saved) {
-      consola.success(`Export file "${template?.name}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name_nl}" imported from ${type} ${envId}`);
     }
   } catch (error) {
     consola.error(error);
@@ -258,7 +258,7 @@ async function fetchAllExportFiles(type, envId, page = 1) {
     const saved = ExportFile.save(type, envId, template);
 
     if (saved) {
-      consola.success(`Export file "${template?.name}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name_nl}" imported from ${type} ${envId}`);
     }
   });
   fetchAllExportFiles(type, envId, page + 1);
--- a/index.js
+++ b/index.js
@@ -211,7 +211,7 @@ async function fetchExportFileByName(type, envId, name) {
 
     const saved = ExportFile.save(type, envId, template);
     if (saved) {
-      consola.success(`Export file "${name}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name_nl}" imported from ${type} ${envId}`);
     }
   } catch (error) {
     consola.error(error);
@@ -258,7 +258,7 @@ async function fetchAllExportFiles(type, envId, page = 1) {
     const saved = ExportFile.save(type, envId, template);
 
     if (saved) {
-      consola.success(`Export file "${template.name}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name_nl}" imported from ${type} ${envId}`);
     }
   });
   fetchAllExportFiles(type, envId, page + 1);
@@ -315,8 +315,8 @@ async function publishExportFileByName(type, envId, name, message = "updated thr
 
     const response = await SF.updateExportFile(type, envId, templateId, template);
 
-    if (response && response.data && response.data.name) {
-      consola.success(`Export file updated: ${response.data.name}`);
+    if (response && response.data && response.data.name_nl) {
+      consola.success(`Export file updated: ${response.data.name_nl}`);
       return true;
     } else {
       consola.error(`Export file update failed: ${name}`);
@@ -346,11 +346,11 @@ async function newExportFile(type, envId, name) {
     if (!template) return;
     template.version_comment = "Created through the Silverfin CLI";
     const response = await SF.createExportFile(type, envId, template);
-
+    const handle = response.data.name_nl;
     // Store new id
     if (response && response.status == 201) {
-      ExportFile.updateTemplateId(type, envId, name, response.data.id);
-      consola.success(`Export file "${name}" created on ${type} ${envId}`);
+      ExportFile.updateTemplateId(type, envId, handle, response.data.id);
+      consola.success(`Export file "${handle}" created on ${type} ${envId}`);
     }
   } catch (error) {
     errorUtils.errorHandler(error);
--- a/index.js
+++ b/index.js
@@ -211,7 +211,7 @@ async function fetchExportFileByName(type, envId, name) {
 
     const saved = ExportFile.save(type, envId, template);
     if (saved) {
-      consola.success(`Export file "${template?.name_nl}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name}" imported from ${type} ${envId}`);
     }
   } catch (error) {
     consola.error(error);
@@ -258,7 +258,7 @@ async function fetchAllExportFiles(type, envId, page = 1) {
     const saved = ExportFile.save(type, envId, template);
 
     if (saved) {
-      consola.success(`Export file "${template?.name_nl}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name}" imported from ${type} ${envId}`);
     }
   });
   fetchAllExportFiles(type, envId, page + 1);
--- a/lib/templates/accountTemplate.js
+++ b/lib/templates/accountTemplate.js
@@ -16,14 +16,8 @@ class AccountTemplate {
    * @param {object} template The object to be processed and saved
    */
   static save(type, envId, template) {
-    if (!template?.name_nl) {
-      consola.warn(
-        `Template name_nl is missing "${
-          template?.name_en || template?.name_fr || template?.name_da || template?.name_de
-        }". Skipping. NL must be enabled in "Advanced Settings" in Silverfin because the NL name is the only required field for a template name.`
-      );
-      return false;
-    } // NL must be enabled in "Advanced Settings" in Silverfin
+    // NL must be enabled in "Advanced Settings" in Silverfin
+    if (templateUtils.missingNameNL(template)) return false;
     if (templateUtils.missingLiquidCode(template)) return false;
     if (!templateUtils.checkValidName(template?.name_nl, this.TEMPLATE_TYPE)) return false;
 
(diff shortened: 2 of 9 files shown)


AgustinSilverfin:
@michieldegezelle 

> * Do we still need the 'name' attribute in the config in the ExportFile class?

I'm of the opinion that we should not keep name. As soon as name_en/fr/nl are added, we should also remove name. That way they never live together.

> * Do we want to remove the 'name' attribute from the config files as soon as someone does a 'import-export-file'? Now, we only add the 3 new locales.

I would also do it when importing. Of course, until it's updated it will be correct in the repo but not yet in Silverfin.
I don't see major issues with that?
And it will also help with your following topic of changes after updated (you will have your names already correct in your PR)

> * When doing an update via 'update-export-file', some new logic was added to adjust the config file and add the missing name translations. This leads to 'new' changes after we do the update command (which can in some cases e.g. update to partners be the last action). As a result, we can end up with a situation where a fix was approved, merged with main, and when someone wants to push from main with the update command, we add the new locales. These new changes need to be put in a new PR to be reflected correctly in the repo's. Do you see any issues with this? This is a one-off of course.

Yes, a bit annoying. Including it in import as suggested earlier may help.
I don't see another issues, If possible evaluate to update all export files at once in one PR, and that's all. Hopefully there are not to many export files.

> * Should we consider renaming the folder automatically to the nl naming as soon as we do any updates?

Let's discuss this and pick it up as a follow up. Since it could be valid for any template type.

> In the update command, we refer to the new nl name, but the folder name is still referring to the old name. We end up in the same situation as when we have a typo: Warning about ID that is missing.
> Given that this is a one-off, instead of having to write some additional logic here, could be maybe make the warning a little bit more descriptive?
> So: ID is missing, aborted. Please check your command for typos and check if the folder name matches the name_nl (or something like this).

Yes, I like the  idea of extending the working to check typos and folder names

michieldegezelle:
> @michieldegezelle
> 
> > * Do we still need the 'name' attribute in the config in the ExportFile class?
> 
> I'm of the opinion that we should not keep name. As soon as name_en/fr/nl are added, we should also remove name. That way they never live together.

**Ok, so removing it here shouldn't give any issues? (as long as we cover the conversion in all cases)**

image


> 
> > * Do we want to remove the 'name' attribute from the config files as soon as someone does a 'import-export-file'? Now, we only add the 3 new locales.
> 
> I would also do it when importing. Of course, until it's updated it will be correct in the repo but not yet in Silverfin. I don't see major issues with that? And it will also help with your following topic of changes after updated (you will have your names already correct in your PR)

**Alright, I will try to add that logic as well**

> > * When doing an update via 'update-export-file', some new logic was added to adjust the config file and add the missing name translations. This leads to 'new' changes after we do the update command (which can in some cases e.g. update to partners be the last action). As a result, we can end up with a situation where a fix was approved, merged with main, and when someone wants to push from main with the update command, we add the new locales. These new changes need to be put in a new PR to be reflected correctly in the repo's. Do you see any issues with this? This is a one-off of course.
> 
> Yes, a bit annoying. Including it in import as suggested earlier may help. I don't see another issues, If possible evaluate to update all export files at once in one PR, and that's all. Hopefully there are not to many export files.

**Luckily, the number of export files is limited. We could indeed make sure that every market creates 1 PR where they fix this all together. Only partners could then have some issues with this.
Adding it to the import-commands might already cover a couple of cases, but in most cases (at least in BE), we don't import before we start developing. We just start from main. This issue would thus still pop up in that flow**

> > * Should we consider renaming the folder automatically to the nl naming as soon as we do any updates?
> 
> Let's discuss this and pick it up as a follow up. Since it could be valid for any template type.

**Alright, will leave it out of scope for now. **

> > In the update command, we refer to the new nl name, but the folder name is still referring to the old name. We end up in the same situation as when we have a typo: Warning about ID that is missing.
> > Given that this is a one-off, instead of having to write some additional logic here, could be maybe make the warning a little bit more descriptive?
> > So: ID is missing, aborted. Please check your command for typos and check if the folder name matches the name_nl (or something like this).

> 
> Yes, I like the idea of extending the working to check typos and folder names

**I was mainly referring to the warning text, not really to the logic (yet). Would adjusting the warning be enough for now? **

AgustinSilverfin:
@michieldegezelle  Now that I think about it a bit more, import should already return the new ones. Since Silverfin no longer includes `name` and it will give you the present locales. So import I would just behave as Account Templates, and remove existing name also, yes.

AgustinSilverfin (commented):
Changes look good in general. 
We need someone to test it functionally, first. 
And we should discuss what we need to do around naming in general as a follow up

index.js
```diff
@@ -236,15 +236,15 @@ async function newAllReconciliations(type, envId) {
...
+
     if (!template) {
       consola.error(`Export file "${name}" wasn't found in ${type} ${envId}`);
       process.exit(1);
     }
 
     const saved = ExportFile.save(type, envId, template);
     if (saved) {
-      consola.success(`Export file "${name}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name_nl}" imported from ${type} ${envId}`);
```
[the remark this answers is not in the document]
delphine-demeulenaere: Adjusted!

lib/templates/accountTemplate.js
```diff
@@ -16,7 +16,7 @@ class AccountTemplate {
   ];
   static TEMPLATE_TYPE = "accountTemplate";
   static TEMPLATE_FOLDER = fsUtils.FOLDERS[this.TEMPLATE_TYPE];
-  constructor() {}
+  constructor() { }
```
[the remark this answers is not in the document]
delphine-demeulenaere: Adjusted! I installed prettier, thanks for the tip!

lib/templates/exportFile.js
```diff
@@ -62,9 +69,22 @@ class ExportFile {
...
+    // Handle legacy name conversion
+    if (templateConfig.name && !templateConfig.name_nl) {
+      templateConfig.name_nl = templateConfig.name;
+      templateConfig.name_en = templateConfig.name;
+      templateConfig.name_fr = templateConfig.name;
+      delete templateConfig.name;
+
+      // Write updated config back to file
+      fsUtils.writeConfig(this.TEMPLATE_TYPE, name, templateConfig);
+    }
```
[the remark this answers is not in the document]
delphine-demeulenaere: Adjusted!

michieldegezelle:
@AgustinSilverfin This can be reviewed again

index.js
```diff
@@ -211,7 +211,7 @@ async function fetchExportFileByName(type, envId, name) {
 
     const saved = ExportFile.save(type, envId, template);
     if (saved) {
-      consola.success(`Export file "${name}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name}" imported from ${type} ${envId}`);
```
AgustinSilverfin, line 214: I think it should be `template.name_nl`, since `.name` will no longer exist

index.js
```diff
@@ -258,7 +258,7 @@ async function fetchAllExportFiles(type, envId, page = 1) {
     const saved = ExportFile.save(type, envId, template);
 
     if (saved) {
-      consola.success(`Export file "${template.name}" imported from ${type} ${envId}`);
+      consola.success(`Export file "${template?.name}" imported from ${type} ${envId}`);
```
AgustinSilverfin, line 261: I think it should be `template.name_nl`, since `.name` will no longer exist

lib/templates/exportFile.js
```diff
@@ -53,9 +56,17 @@ class ExportFile {
...
-    const templateConfig = fsUtils.readConfig(this.TEMPLATE_TYPE, name);
+    let templateConfig = fsUtils.readConfig(this.TEMPLATE_TYPE, name);
+
+    // Handle legacy name conversion
+    templateConfig = this.#populateMissingNameLocales(templateConfig); 
+
+    // Write updated config back to file
+    fsUtils.writeConfig(this.TEMPLATE_TYPE, name, templateConfig);
+
+    if (!this.#validateFolderName(name, templateConfig)) return false;
```
AgustinSilverfin, line 67: :+1:

AgustinSilverfin: Maybe we should implement the same for Account Templates if it's not there yet

AgustinSilverfin (approved):
Left some minor comments but In general looks good

index.js
delphine-demeulenaere: Adjusted!

index.js
delphine-demeulenaere: Adjusted!

lib/templates/exportFile.js
```diff
@@ -117,6 +114,10 @@ class ExportFile {
       templateConfig.name_en = templateConfig.name;
       templateConfig.name_fr = templateConfig.name;
       delete templateConfig.name;
+
+      // Write updated config back to file
+      consola.debug(`Template ${templateConfig.name_nl}: adjusting name fields in config.json`);
+      fsUtils.writeConfig(this.TEMPLATE_TYPE, templateConfig.name_nl, templateConfig);
```
AgustinSilverfin, line 120: I moved `writeConfig` to avoid calling it every time we call `read` (for example, every time you use an `update-` command). So in this case, we only write it when we are updating the names in it.
See the screenshot, the first time I run the command, we update names and write. The second time there is no need to update or write

image

PR #182, opened 2025-05-16
Validate cwd when running Silverfin CLI

### LINKED ISSUE #71: CLI: Check CWD before running commands

The CLI is intend to be used from the root directory of your repository. If not, it could lead to unexpected results. We should add some checks of the CWD before running any command.

The CLI won't run any command if it is not executed in the right directory. Even better, it will automatically switch the CWD to the correct one.

There are two scenarios:
- Running the CLI command from a directory inside our repository (e.g. /reconciliations_texts/template_handle): In this case, we could identify the root of the repo and switch the CWD before running any command.
- Running the CLI command from a directory outisde our repository: In this case, how can we identify that we are not placed in the right repository ?
  - Checking the existence of folders like `reconciliation_texts` ? What if it's a new repository ?
  - Checking the existence of a git repository ?
  - Should we create some kind of configuration file at the root of the folder that we can search for ?
  - I would use a combination of the first two: if there are no related folders, don't prevent the run of the commands but show a warning. If there is no git repository, prevent the run and show the error, asking to move to the right path. 

Originally raised by @MelSFNL

AgustinSilverfin:
## Description

We want to identify if the program was called from a different place than the root directory of the repository.

If we identify any of the known folders in the path, it means that we are not set in the root, and we will warn it (eventually, we could set the cwd to the root path)

Also, we don't find a git repository we would warn it as well as it may be another sign of a wrong execution

Fixes #71 

## Type of change

- [ ] Bug fix
- [X] New feature
- [ ] Breaking change

## Checklist

### DIFF
--- a/lib/cli/cwdValidator.js
+++ b/lib/cli/cwdValidator.js
@@ -1,0 +1,33 @@
+const { consola } = require("consola");
+const path = require("path");
+const fs = require("fs");
+const fsUtils = require("../utils/fsUtils");
+
+class CwdValidator {
+  static run() {
+    const currentDirectory = process.cwd();
+    const warningMessage = `Please, double check that you are executing "silverfin" CLI in the correct directory. Your current directory is "${currentDirectory}".`;
+    const gitDirPresent = fs.existsSync(path.join(currentDirectory, `.git`));
+
+    if (gitDirPresent) {
+      return;
+    }
+
+    let folderIdentified = false;
+    const folders = Object.values(fsUtils.FOLDERS);
+    for (const folder of folders) {
+      const re = new RegExp(folder);
+
+      if (re.test(currentDirectory)) {
+        folderIdentified = true;
+        consola.warn(`${warningMessage} You are running "silverfin" from the "${folder}" directory, this could have unexpected consequences.`);
+      }
+    }
+
+    if (!folderIdentified) {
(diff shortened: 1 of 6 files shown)


lib/cli/cwdValidator.js
```diff
@@ -0,0 +1,33 @@
...
+
+class CwdValidator {
+  static validate() {
+    const warningMessage = `Please, double check that you are executing "silverfin" CLI in the correct directory.`;
+    const currentDirectory = process.cwd();
+    const gitDirPresent = fs.existsSync(path.join(currentDirectory, `.git`));
+
+    if (gitDirPresent) {
+      return;
+    }
```
AgustinSilverfin, line 14: Should we check for git repository presence, or are there cases where we use it outside any repo? :thinking:

michieldegezelle (changes requested):
Two inline comments + could you also adjust the CLI version?

lib/cli/cwdValidator.js
AgustinSilverfin: I don't want to be more strict at this point as we may be blocking uses cases that we are not foreseeing now.
A warning should be more than enough for the user to correct their actions in my opinion. And we can always modify it to be more strict if we see the reason to do so in the future.

PR #191, opened 2025-06-05
Add changelog

michieldegezelle:

## Description

Added a changelog in two ways:
- Created a changelog.md file to keep track of the full list of changes.
- Made sure that the latest 3 changes are shown if an update is available. (in the terminal).

Also updated the Github actions to check if the changelog was updated. 

## Testing Instructions

Steps:

1. Run the CLI locally (file path to the changelog should be adjusted). 
`const changelogPath = path.join(__dirname, "../../CHANGELOG.md");` instead of the path to Github. 

2. Not sure if we can test the Github action in advance? 

## Author Checklist

## Reviewer Checklist

- [x] PR has a clear title and description
- [x] Manually tested the changes that the PR introduces
- [x] Changes introduced are covered by tests of acceptable quality

### DIFF
--- a/lib/cli/cliUpdater.js
+++ b/lib/cli/cliUpdater.js
@@ -32,17 +32,18 @@ class CliUpdater {
     try {
       consola.log(`Running command: ${chalk.italic(this.#UPDATE_COMMAND)}`);
 
-      const updateOutput = await exec(this.#UPDATE_COMMAND);
-      const updatedPkgVersion = await exec(this.#VERSION_COMMAND);
       const currentVersion = pkg.version;
       const latestVersion = await this.#getLatestVersion();
-      const recentChanges = await ChangelogReader.fetchChanges(currentVersion, latestVersion);
+
+      const updateOutput = await exec(this.#UPDATE_COMMAND);
+      const updatedPkgVersion = await exec(this.#VERSION_COMMAND);
 
       consola.log(`--------------`);
       consola.log(updateOutput.stdout);
       consola.log(`--------------`);
       consola.success(chalk.bold(`Silverfin CLI succesfully updated to version ${updatedPkgVersion.stdout}`));
 
+      const recentChanges = await ChangelogReader.fetchChanges(currentVersion, latestVersion);
       if (recentChanges) {
         consola.log(chalk.underline(`Changes since last update:\n`));
         consola.log(`${recentChanges}\n`);
--- a/.github/workflows/cli_version.yml
+++ b/.github/workflows/cli_version.yml
@@ -77,10 +77,10 @@ jobs:
         id: changelog-check
         run: |
           current_version=$(jq -r '.version' current-branch/package.json)
-          changelog_file="CHANGELOG.md"
+          changelog_file="current-branch/CHANGELOG.md"
           version_entry="## [$current_version]"
 
-          if ! grep -q "$version_entry" "$changelog_file"; then
-            echo "Error: Changelog entry for version $current_version is missing in $changelog_file"
-            exit 1
+          if ! grep -Fq "$version_entry" "$changelog_file"; then
+              echo "Error: Changelog entry for version $current_version is missing in $changelog_file"
+              exit 1
           fi


.github/workflows/changelog-check.yml
AgustinSilverfin, line 1: I don't think we need a separate github action here.
If we do want to have an action that checks the changelog, we should unify it with the [cli version action](https://github.com/silverfin/sf-toolkit/blob/f81f6e262bfad5b5c1c74f6320e1cb68ede468bf/.github/workflows/cli_version.yml#L1) (we can add a step there)

lib/cli/cliUpdater.js
```diff
@@ -68,10 +76,43 @@ class CliUpdater {
...
+        const changeContent = version.split(")")[1];
+        if (changeContent) {
+          changes += changeContent.trim() + "\n";
+        }
+      }
+
+      return changes;
+    } catch (err) {
+      return null;
     }
```
AgustinSilverfin, line 116: Some personal opinions here:

I think this could be a good candidate to have a separate service class. Something that we can call as, for example,  `ChangelogReader.fetchChanges(previousVersion, currentVersion)` here. (names of the class or method can be different :smile: )

Considering that we know current and previous version I would try to display entries between them instead of a fix amount of entries.

We should also add tests now that we are introducing it.

lib/cli/cliUpdater.js
```diff
@@ -6,6 +6,7 @@ const { consola } = require("consola");
 
 const pkg = require("../../package.json");
 const PACKAGE_URL = "https://raw.githubusercontent.com/silverfin/silverfin-cli/main/package.json";
+const CHANGELOG_URL = "https://raw.githubusercontent.com/silverfin/silverfin-cli/main/CHANGELOG.md";
```
AgustinSilverfin, line 9: This constants could be both moved to the class itself, similar to `static #UPDATE_COMMAND or `VERSION_COMMAND`

.github/workflows/changelog-check.yml
michieldegezelle: I will leave it out for now then, we can always add it later on if needed

tests/lib/cli/cliUpdater.spec.js
```diff
@@ -60,7 +60,7 @@ describe("CliUpdater", () => {
       await CliUpdater.checkVersions();
 
       expect(consola.log).not.toHaveBeenCalled();
-      expect(consola.debug).toHaveBeenCalledWith(expect.stringMatching(/Failed to get the latest version/i));
+      expect(consola.debug).not.toHaveBeenCalled();
```
michieldegezelle, line 63: @AgustinSilverfin I had to change this, but don't really know why it was failing.

michieldegezelle: Re-instated and fixed the test again

lib/cli/changelogReader.js
```diff
@@ -0,0 +1,64 @@
...
+          break;
+        } else if (foundUpdateVersion) {
+          relevantSections.push(versionContent);
+        }
+      }
+
+      if (!foundUpdateVersion) {
+        consola.warn(
+          `Version ${updateVersion} not found in the changelog. This might indicate that the changelog hasn't been updated yet for this version, or the version format is incorrect (should be ## [version] (date)).`
+        );
```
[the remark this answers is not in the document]
michieldegezelle: Also updated the tests for this change

lib/cli/cliUpdater.js
```diff
@@ -33,11 +34,16 @@ class CliUpdater {
...
+      const currentVersion = pkg.version;
+      const latestVersion = await this.#getLatestVersion();
+      const recentChanges = await ChangelogReader.fetchChanges(currentVersion, latestVersion);
 
       consola.log(`--------------`);
       consola.log(updateOutput.stdout);
       consola.log(`--------------`);
       consola.success(chalk.bold(`Silverfin CLI succesfully updated to version ${updatedPkgVersion.stdout}`));
+      consola.log(chalk.underline(`Changes since last update:\n`));
+      consola.log(`${recentChanges}\n`);
```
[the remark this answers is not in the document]
michieldegezelle: Added a check

lib/cli/changelogReader.js
```diff
@@ -8,11 +8,7 @@ class ChangelogReader {
     try {
       const response = await axios.get(this.#CHANGELOG_URL);
       if (response.status !== 200) {
-        if (response.status === 404) {
-          consola.debug(`Changelog file not found (404). The CHANGELOG.md file may have been moved or deleted from the repository.`);
-        } else {
-          consola.debug(`Failed to fetch changelog from GitHub. Server returned status ${response.status}`);
-        }
+        consola.debug(`Changelog file not found. The CHANGELOG.md file may have been moved or deleted from the repository.`);
```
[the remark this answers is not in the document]
michieldegezelle: Applied!

PR #200, opened 2025-07-07
Feature branch - 174 - Template names locales

delphine-demeulenaere:

## Description

This is a feature branch on `174-template-names-support-7-locales`. In `174-template-names-support-7-locales` we added 7 locales, in this branch we made sure that not all locales get populated automatically. 

## Testing Instructions

Steps:

1. ...
2. ...
3. ...

## Author Checklist

## Reviewer Checklist


lib/utils/fsUtils.js
```diff
@@ -161,8 +161,12 @@ function createConfigIfMissing(templateType, handle) {
     case "reconciliationText":
       config.handle = handle;
       config.name_nl = `${handle}`;
-      config.name_fr = `${handle}`;
-      config.name_en = `${handle}`;
+      config.name_fr = ``;
```
michieldegezelle, line 164: Can you use "" instead of the backticks?

michieldegezelle (changes requested):

AgustinSilverfin:
Sorry for the delay on this @delphine-demeulenaere @michieldegezelle 

I did an import and now we will always have all name items in the config json even if not needed? This will introduce many changes in existing repos.

Do you think we should consider filtering this out? For example, only updating the ones that have a content from the API when we import. The question here is if we should remove the ones that we already have but are not present in the latest import (I wouldn't because different firms could have different locale settings)

image

AgustinSilverfin (changes requested):
Changes look good in general, but see my latest commit about filtering further.

delphine-demeulenaere:
> Sorry for the delay on this @delphine-demeulenaere @michieldegezelle
> 
> I did an import and now we will always have all name items in the config json even if not needed? This will introduce many changes in existing repos.
> 
> Do you think we should consider filtering this out? For example, only updating the ones that have a content from the API when we import. The question here is if we should remove the ones that we already have but are not present in the latest import (I wouldn't because different firms could have different locale settings)
>

Hi Rufo! I started looking into your suggestion above, but I'm not entirely sure what the best approach here would be. Could you maybe share how you'd tackle this? Also, this seems more like an edge case, right? We almost always create reconciliations via the CLI in Cursor/VSC, rather than directly through the platform.

AgustinSilverfin (approved):
You will need to address the conflicts before merging.

PR #218, opened 2025-08-29
Provide terminal tab autocompletion

### LINKED ISSUE #216: [Feature Request] Autocompletion
When the user type in the terminal `silverfin update-reconciliation -h exam` and hit `TAB`, the rest of the handle should be auto completed, leading to `silverfin update-reconciliation -h example_handle`

**Additional context**
This should be valid for every template type. The handle should be looked into the corresponding folder, based on existing templates in the repo.
If there is only one matching option, it should use that one. If there are many, none will be used and all possible options should be displayed as a guide to the user.

AgustinSilverfin:
Fixes #216 

## Description

The goal is to provide tab autocompletions for `silverfin` command in the terminal. It should auto complete command names, flag names for that specific command, and template names from the current directory.

## Testing Instructions

Steps:

1. Run `silverfin config --set-autocompletion`
2. Follow the instructions, adding the `source` line in your `~/.zshrc` or `~/.bashrc` file
3. Source your file to apply changes in your current session (e.g. `source ~/.zshrc`)
4. Test the auto completion in a template repository.

## Author Checklist

## Reviewer Checklist


BenjaminLangenakenSF (changes requested):
Works as intended for `reconciliation_texts` and `shared_parts` 👌
However, for `account_templates` and `export_files` (which use the template name), the autocomplete doesn’t seem to work when there’s whitespace between the words?
e.g:
Image

AgustinSilverfin:
> Works as intended for `reconciliation_texts` and `shared_parts` 👌 However, for `account_templates` and `export_files` (which use the template name), the autocomplete doesn’t seem to work when there’s whitespace between the words? e.g: Image

Thanks @BenjaminLangenakenSF it should be solved now! could you give it another try? :slightly_smiling_face:

BenjaminLangenakenSF (changes requested):
Seems the autocomplete now always refers to the repo folders instead of the CLI commands?
Image

michieldegezelle (commented):
I did a couple of tests, and did have the issue that Benji encountered, so all good. 

Couple of remarks:
- If I type `silverfin import-reconciliation --handle <TAB>`, I get an error:
image

-> Is this to be expected? The list is probably too long to display, but we might have to avoid the error?

- Shortened options are not autocompleted, can this easily be added (e.g. -h or -t etc.)? 

- Seems to be for all 'second' options: 
image

-> Wanted to add the firm here, and got the error. Could be that I am just willing to autocomplete too much

Apart from this, all good! 👍

PR #221, opened 2025-09-10
feat: new command to generate export file instances

AgustinSilverfin:

## Description

## Testing Instructions

Steps:

1. ...
2. ...
3. ...

## Author Checklist

## Reviewer Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -3,6 +3,7 @@
 const toolkit = require("../index");
 const liquidTestGenerator = require("../lib/liquidTestGenerator");
 const liquidTestRunner = require("../lib/liquidTestRunner");
+const { ExportFileInstanceGenerator } = require("../lib/exportFileInstanceGenerator");
 const stats = require("../lib/cli/stats");
 const { Command, Option } = require("commander");
 const pkg = require("../package.json");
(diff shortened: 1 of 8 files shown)


bin/cli.js
```diff
@@ -650,6 +650,29 @@ program
     }
   });
 
+// GENERATE export file
+program
+  .command("generate-export-file")
```
michieldegezelle: This seems to work, but it created no file name. So I just get <blank>.xml.

Should we give it a default name?

bin/cli.js
```diff
@@ -650,6 +650,29 @@ program
...
+  .description("Create a new export file instance from an export file template")
+  .option("-f, --firm <firm-id>", "Specify the firm to be used", firmIdDefault)
+  .requiredOption("-c, --company <company-id>", "Specify the company to be used")
+  .requiredOption("-p, --period <period-id>", "Specify the period to be used")
+  .requiredOption("-e, --export-file <export-file-id>", "Specify the export file template to be used")
+  .option("--url", "Get the url of the generated export file (optional)", false)
```
michieldegezelle: Should we make this the default?

index.js
```diff
@@ -1166,6 +1166,52 @@ async function updateFirmName(firmId) {
...
+          `Export file generation completed for firm ${firmId}, company ${companyId}, period ${periodId}, export file ${exportFileId}. Instance ID: ${exportFileInstanceId}`
+        );
+        return responseFetch;
+      } else {
+        consola.error(
+          `Export file generation encountered an unexpected state for firm ${firmId}, company ${companyId}, period ${periodId}, export file ${exportFileId}. Instance ID: ${exportFileInstanceId}. State: ${responseFetch ? responseFetch.state : "unknown"}`
+        );
+        return false;
+      }
+    }
```
michieldegezelle: Should we add a return for when the max attempts are reached?

bin/cli.js
AgustinSilverfin: Yes, I have remove the --url flag for now, and it should open the export file in the browser by default. We could modify if we ever see any other use cases but for now I assume that if you want to generate an export, is because you want to open the provided file

index.js
AgustinSilverfin: 👍

bin/cli.js
```diff
@@ -650,6 +650,29 @@ program
     }
   });
 
+// GENERATE export file
+program
+  .command("generate-export-file")
+  .description("Create a new export file instance from an export file template")
+  .option("-f, --firm <firm-id>", "Specify the firm to be used", firmIdDefault)
+  .requiredOption("-c, --company <company-id>", "Specify the company to be used")
+  .requiredOption("-p, --period <period-id>", "Specify the period to be used")
```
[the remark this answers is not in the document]
michieldegezelle: @AgustinSilverfin Is this something that we should take into account? 

CodeRabbit
Consider the -p flag inconsistency with other commands.

The -p flag is used for --period in this command, but in other commands like import-reconciliation, -p is used for --partner. While Commander.js handles this correctly per-command, users familiar with other CLI commands might expect -p to mean --partner.

Consider using a different short flag for period (e.g., -P or -d for "date period") to maintain consistency across commands.

PR #233, opened 2025-12-03
Run multiple handles in parallel

michieldegezelle:

## Description

Multiple handles/AT's can now be passed to the run-test command:

e.g. `silverfin run-test -f 200619 -h be_pit_wages_salaries_part_2 be_pit_wages_salaries --status`
or  `silverfin run-test -f 14400 -h 'Dubieuze debiteuren' 'Erelonen' --status`

Note that I only allowed/activated it for the status flag since this will cover our usecase for now (Github actions). 
This can be expanded later on if required.

Had some issues with the handle flag being rendered as an array in case there was only 1 element/handle, but some checks solved this issue. 

Reporting is done per handle (and if differences, we show the units/tests as well). 

QQ: in general, this is quicker than running them one by one, but in some cases, the API seems to be very slow. Is there a way in which we can avoid that? 

## Testing Instructions

Steps:

1. Run tests with multiple handles
2. Run tests with one handle
3. Run test without staus flag

## Author Checklist

## Reviewer Checklist

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -467,24 +467,20 @@ program
 
     const templateType = options.handle ? "reconciliationText" : "accountTemplate";
     const templateName = options.handle ? options.handle : options.accountTemplate;
-
-    // Normalize to array for uniform handling (Commander returns array for variadic, string otherwise)
-    const templateNames = Array.isArray(templateName) ? templateName : [templateName];
-
     // Block multiple handles/templates without --status
-    if (templateNames.length > 1 && !options.status) {
+    if (templateName.length > 1 && !options.status) {
       consola.error("Multiple handles/templates are only allowed when used with the --status flag");
       process.exit(1);
     }
 
     if (options.status) {
       // Status mode: allow multiple, pass array of template names
-      await liquidTestRunner.runTestsStatusOnly(options.firm, templateType, templateNames, options.test);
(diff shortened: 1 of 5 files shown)


bin/cli.js
```diff
@@ -469,15 +468,30 @@ program
...
-        process.exit(1);
-      }
-      liquidTestRunner.runTestsWithOutput(options.firm, templateType, templateName, options.test, options.previewOnly, options.htmlInput, options.htmlPreview);
+      // Status mode: allow multiple, pass array of template names
+      await liquidTestRunner.runTestsStatusOnly(options.firm, templateType, templateNames, options.test);
+      return;
     }
+
+    // Non-status mode: always run a single template, pass string handle/name
+    const singleTemplateName = templateNames[0];
```
AgustinSilverfin: nitpick, you can use `const templateName` which still exists?

lib/liquidTestRunner.js
AgustinSilverfin: I would personally rename the parameter as `handles` as it is an array already (or can be an array)
Something that gives already a clue that it can be an array

AgustinSilverfin (approved):
I haven't tested it myself but in general it looks good 👍

bin/cli.js
michieldegezelle: image

PR #234, opened 2025-12-17
Create tests for account templates

michieldegezelle:

## Description

Make sure that the create test functionality works for account templates as well. 

Limitations: functionality is limited to fetching the template data/period data/custom data. We won't check for template dependencies for now since this is rather rare. 

If needed, that can be added later on. 

## Testing Instructions

Steps:

1. Test a random account template with lots of data
2. Include a period.custom drop + company drop in the code to see if it is fecthed
3. Check if it is stored in the correct folder/path
4. Check if the logic still works for reconciliation templates

## Author Checklist

## Reviewer Checklist


lib/liquidTestGenerator.js
```diff
@@ -22,14 +24,50 @@ async function testGenerator(url, testName, reconciledStatus = true) {
...
+      }
+      break;
+    }
+  }
 
   // Get Workflow Information
-  const starredStatus = {
-    ...SF.findReconciliationInWorkflow(parameters.firmId, reconciliationHandle, parameters.companyId, parameters.ledgerId, parameters.workflowId),
-  }.starred;
+  let starredStatus;
```
BenjaminLangenakenSF, line 60: I checked if the starredStatus status flows through to the generated YAML files for AT's and RT's when using create test.
And seems this isn't the case?
(also checked this on the `main` branch, and seems there it also doesn't seem to work?)

BenjaminLangenakenSF (changes requested):
Checked functionally, and the liquid test YAML file for ATs is being generated correctly 💪  
I do have one question related to the starred status, could you have a look?
And feel free to reach out if you want to discuss 😉

lib/liquidTestGenerator.js
michieldegezelle: Well spotted. For reconciliations, the function call needed to be async (so added await).

For account templates, starred status was not logged/added. Both are solved now

lib/liquidTestGenerator.js
```diff
@@ -57,125 +95,159 @@ async function testGenerator(url, testName, reconciledStatus = true) {
...
+      templateCode = await ReconciliationText.read(templateHandle);
+      break;
+    case "accountTemplate":
+      templateCode = await AccountTemplate.read(templateHandle);
+      break;
+  }
+
+  if (!templateCode) {
+    consola.warn(`Template "${templateHandle}" wasn't found`);
     process.exit();
```
BenjaminLangenakenSF: I think coderabbit is correct, and we should also use `process.exit(1);` here?

lib/liquidTestGenerator.js
```diff
@@ -57,125 +95,159 @@ async function testGenerator(url, testName, reconciledStatus = true) {
...
   }
```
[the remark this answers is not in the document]
BenjaminLangenakenSF: Added review comment for this issue

BenjaminLangenakenSF (approved):
Added one final comment, but will already approve the PR 😉

lib/liquidTestGenerator.js
BenjaminLangenakenSF: Tested this functionally and seems we indeed still create an empty liquid test file in case of `!templateCode`:
image

michieldegezelle: Adjusted!

PR #236, opened 2026-01-07
feat: liquid sampler command

AgustinSilverfin:

## Description

## Testing Instructions

Steps:

1. ...
2. ...
3. ...

## Author Checklist

## Reviewer Checklist

- [x] PR has a clear title and description
- [x] Manually tested the changes that the PR introduces

### GH Action
[GH action](https://github.com/silverfin/be_market/blob/ad16b97a0d26b006a6c6b47965d2237de387b2be/.github/workflows/run_sampler_TEST.yml) currently calls run-sampler as follows:  `run-sampler -p ${SAMPLER_PARTNER} -h ${SAMPLER_HANDLES} ${FIRM_ARGS} --compact`
 - `--compact` added as part of 2nd PR atop this PR
 - Variable `FIRM_ARGS` is simply `-f` option followed by a list of firms

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -4,6 +4,7 @@ const toolkit = require("../index");
 const liquidTestGenerator = require("../lib/liquidTestGenerator");
 const liquidTestRunner = require("../lib/liquidTestRunner");
 const { ExportFileInstanceGenerator } = require("../lib/exportFileInstanceGenerator");
+const { LiquidSamplerRunner } = require("../lib/liquidSamplerRunner");
 const stats = require("../lib/cli/stats");
 const { Command, Option } = require("commander");
 const pkg = require("../package.json");
@@ -512,6 +513,44 @@ program
       options.htmlPreview,
       options.pattern
     );
+  });
+
+// Run Liquid Sampler
+program
+  .command("run-sampler")
+  .description("Run Liquid Sampler for partner templates (reconciliation texts, account detail templates, and/or shared parts)")
+  .requiredOption("-p, --partner <partner-id>", "Specify the partner to be used")
+  .option("-h, --handle <handles...>", "Specify reconciliation text handle(s) - can specify multiple")
+  .option("-at, --account-template <names...>", "Specify account detail template name(s) - can specify multiple")
+  .option("-s, --shared-part <names...>", "Specify shared part name(s) - can specify multiple")
+  .option("--firm-ids <firm-ids...>", "Specify firm ID(s) to run the sampler against - can specify multiple (optional)")
+  .option("--id <sampler-id>", "Specify an existing sampler ID to fetch results for (optional)")
+  .action(async (options) => {
+    // If an existing sampler ID is provided, fetch and display results
+    if (options.id) {
+      await new LiquidSamplerRunner(options.partner).checkStatus(options.id);
+      return;
+    }
+
+    // Validate: at least one template specified
+    const handles = options.handle || [];
+    const accountTemplates = options.accountTemplate || [];
+    const sharedParts = options.sharedPart || [];
+
+    if (handles.length === 0 && accountTemplates.length === 0 && sharedParts.length === 0) {
+      consola.error("You need to specify at least one template using -h, -at, or -s");
+      process.exit(1);
+    }
+
+    const templateHandles = {
+      reconciliationTexts: handles,
+      accountTemplates: accountTemplates,
+      sharedParts: sharedParts,
+    };
+
+    const firmIds = options.firmIds || [];
+
+    await new LiquidSamplerRunner(options.partner).run(templateHandles, firmIds);
   });
 
 // Create Liquid Test
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -523,7 +523,8 @@ program
   .option("-h, --handle <handles...>", "Specify reconciliation text handle(s) - can specify multiple")
   .option("-at, --account-template <names...>", "Specify account detail template name(s) - can specify multiple")
   .option("-s, --shared-part <names...>", "Specify shared part name(s) - can specify multiple")
-  .option("-i, --id <sampler-id>", "Specify an existing sampler ID to fetch results for (optional)")
+  .option("--firm-ids <firm-ids...>", "Specify firm ID(s) to run the sampler against - can specify multiple (optional)")
+  .option("--id <sampler-id>", "Specify an existing sampler ID to fetch results for (optional)")
   .action(async (options) => {
     // If an existing sampler ID is provided, fetch and display results
     if (options.id) {
@@ -547,7 +548,9 @@ program
       sharedParts: sharedParts,
     };
 
-    await new LiquidSamplerRunner(options.partner).run(templateHandles);
+    const firmIds = options.firmIds || [];
+
+    await new LiquidSamplerRunner(options.partner).run(templateHandles, firmIds);
   });
 
 // Create Liquid Test
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -4,6 +4,7 @@ const toolkit = require("../index");
 const liquidTestGenerator = require("../lib/liquidTestGenerator");
 const liquidTestRunner = require("../lib/liquidTestRunner");
 const { ExportFileInstanceGenerator } = require("../lib/exportFileInstanceGenerator");
+const { LiquidSamplerRunner } = require("../lib/liquidSamplerRunner");
 const stats = require("../lib/cli/stats");
 const { Command, Option } = require("commander");
 const pkg = require("../package.json");
@@ -512,6 +513,41 @@ program
       options.htmlPreview,
       options.pattern
     );
+  });
+
+// Run Liquid Sampler
+program
+  .command("run-sampler")
+  .description("Run Liquid Sampler for partner templates (reconciliation texts, account detail templates, and/or shared parts)")
+  .requiredOption("-p, --partner <partner-id>", "Specify the partner to be used")
+  .option("-h, --handle <handles...>", "Specify reconciliation text handle(s) - can specify multiple")
+  .option("-at, --account-template <names...>", "Specify account detail template name(s) - can specify multiple")
+  .option("-s, --shared-part <names...>", "Specify shared part name(s) - can specify multiple")
+  .option("-i, --id <sampler-id>", "Specify an existing sampler ID to fetch results for (optional)")
+  .action(async (options) => {
+    // If an existing sampler ID is provided, fetch and display results
+    if (options.id) {
+      await new LiquidSamplerRunner(options.partner).checkStatus(options.id);
+      return;
+    }
+
+    // Validate: at least one template specified
+    const handles = options.handle || [];
+    const accountTemplates = options.accountTemplate || [];
+    const sharedParts = options.sharedPart || [];
+
+    if (handles.length === 0 && accountTemplates.length === 0 && sharedParts.length === 0) {
+      consola.error("You need to specify at least one template using -h, -at, or -s");
+      process.exit(1);
+    }
+
+    const templateHandles = {
+      reconciliationTexts: handles,
+      accountTemplates: accountTemplates,
+      sharedParts: sharedParts,
+    };
+
+    await new LiquidSamplerRunner(options.partner).run(templateHandles);
   });
 
 // Create Liquid Test
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -517,7 +517,7 @@ program
 
 // Run Liquid Sampler
 program
-  .command("run-sampler")
+  .command("run-sampler", { hidden: true})
   .description("Run Liquid Sampler for partner templates (reconciliation texts, account detail templates, and/or shared parts)")
   .requiredOption("-p, --partner <partner-id>", "Specify the partner to be used")
   .option("-h, --handle <handles...>", "Specify reconciliation text handle(s) - can specify multiple")
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -524,10 +524,21 @@ program
   .option("-at, --account-template <names...>", "Specify account detail template name(s) - can specify multiple")
   .option("-s, --shared-part <names...>", "Specify shared part name(s) - can specify multiple")
   .option("--firm-ids <firm-ids...>", "Specify firm ID(s) to run the sampler against - can specify multiple (optional)")
-  .option("--id <sampler-id>", "Specify an existing sampler ID to fetch results for (optional)")
+  .addOption(
+    new Option("--id <sampler-id>", "Specify an existing sampler ID to fetch results for (optional)").conflicts([
+      "handle",
+      "accountTemplate",
+      "sharedPart",
+      "firmIds",
+    ])
+  )
   .action(async (options) => {
     // If an existing sampler ID is provided, fetch and display results
     if (options.id) {
+      if (!/^\d+$/.test(options.id)) {
+        consola.error("Invalid sampler ID: must be a numeric value");
+        process.exit(1);
+      }
       await new LiquidSamplerRunner(options.partner).checkStatus(options.id);
       return;
     }
--- a/lib/api/sfApi.js
+++ b/lib/api/sfApi.js
@@ -717,6 +717,30 @@ async function getExportFileInstance(firmId, companyId, periodId, exportFileInst
   }
 }
 
+async function createSamplerRun(partnerId, attributes) {
+  const instance = AxiosFactory.createInstance("partner", partnerId);
+  try {
+    const response = await instance.post("liquid_sampler/run", attributes);
+    apiUtils.responseSuccessHandler(response);
+    return response;
+  } catch (error) {
+    const response = await apiUtils.responseErrorHandler(error);
+    return response;
+  }
+}
+
+async function readSamplerRun(partnerId, samplerId) {
+  const instance = AxiosFactory.createInstance("partner", partnerId);
+  try {
+    const response = await instance.get(`liquid_sampler/${samplerId}`);
+    apiUtils.responseSuccessHandler(response);
+    return response;
+  } catch (error) {
+    const response = await apiUtils.responseErrorHandler(error);
+    return response;
+  }
+}
+
 module.exports = {
   authorizeFirm,
   refreshFirmTokens,
@@ -771,4 +795,6 @@ module.exports = {
   getFirmDetails,
   createExportFileInstance,
   getExportFileInstance,
+  createSamplerRun,
+  readSamplerRun,
 };
--- a/lib/liquidSamplerRunner.js
+++ b/lib/liquidSamplerRunner.js
@@ -1,0 +1,247 @@
+const { UrlHandler } = require("./utils/urlHandler");
+const errorUtils = require("./utils/errorUtils");
+const { spinner } = require("./cli/spinner");
+const SF = require("./api/sfApi");
+const fsUtils = require("./utils/fsUtils");
+const { consola } = require("consola");
+
+const { ReconciliationText } = require("./templates/reconciliationText");
+const { AccountTemplate } = require("./templates/accountTemplate");
+const { SharedPart } = require("./templates/sharedPart");
+
+/**
+ * Class to run liquid samplers for partner templates
+ */
+class LiquidSamplerRunner {
+  constructor(partnerId) {
+    this.partnerId = partnerId;
+  }
+
+  /**
+   * Run liquid sampler for partner templates
+   * @param {Object} templateHandles - Object containing arrays of template identifiers
+   * @param {Array<string>} templateHandles.reconciliationTexts - Array of reconciliation text handles
+   * @param {Array<string>} templateHandles.accountTemplates - Array of account template names
+   * @param {Array<string>} templateHandles.sharedParts - Array of shared part names
+   * @returns {Promise<void>}
+   */
+  async run(templateHandles = {}) {
+    try {
+      // Validate at least one template specified
+      const { reconciliationTexts = [], accountTemplates = [], sharedParts = [] } = templateHandles;
+      if (reconciliationTexts.length === 0 && accountTemplates.length === 0 && sharedParts.length === 0) {
+        consola.error("You need to specify at least one template using -h, -at, or -s");
+        process.exit(1);
+      }
+
+      // Build payload
+      const samplerParams = await this.#buildSamplerParams(templateHandles);
+
+      consola.info(`Starting sampler run with ${samplerParams.templates.length} template(s)...`);
+
+      // Start sampler run
+      const samplerResponse = await SF.createSamplerRun(this.partnerId, samplerParams);
+      const samplerId = samplerResponse.data.id || samplerResponse.data;
+
+      if (!samplerId) {
+        consola.error("Failed to start sampler run - no ID returned");
+        process.exit(1);
+      }
+
+      consola.info(`Sampler run started with ID: ${samplerId}`);
+
+      // Poll for completion
+      const samplerRun = await this.#fetchAndWaitSamplerResult(samplerId);
+
+      // Process results
+      await this.#handleSamplerResponse(samplerRun);
+    } catch (error) {
+      errorUtils.errorHandler(error);
+    }
+  }
+
+  /**
+   * Fetch the status of an existing sampler run
+   * @param {string} samplerId - The sampler run ID
+   * @returns {Promise<void>}
+   */
+  async checkStatus(samplerId) {
+    try {
+      consola.info(`Fetching status for sampler run ID: ${samplerId}`);
+
+      const response = await SF.readSamplerRun(this.partnerId, samplerId);
+
+      await this.#handleSamplerResponse(response.data);
+    } catch (error) {
+      errorUtils.errorHandler(error);
+    }
+  }
+
+  /**
+   * Build sampler parameters from local template files
+   * @param {Object} templateHandles - Object containing arrays of template identifiers
+   * @param {Array<string>} templateHandles.reconciliationTexts - Array of reconciliation text handles
+   * @param {Array<string>} templateHandles.accountTemplates - Array of account template names
+   * @param {Array<string>} templateHandles.sharedParts - Array of shared part names
+   * @returns {Object} Sampler payload with templates array
+   */
+  async #buildSamplerParams(templateHandles = {}) {
+    const templates = [];
+    const { reconciliationTexts = [], accountTemplates = [], sharedParts = [] } = templateHandles;
+
+    // Process reconciliation texts
+    for (const handle of reconciliationTexts) {
+      const templateType = "reconciliationText";
+      const configPresent = fsUtils.configExists(templateType, handle);
+
+      if (!configPresent) {
+        consola.error(`Config file for reconciliation text "${handle}" not found`);
+        process.exit(1);
+      }
+
+      const config = fsUtils.readConfig(templateType, handle);
+
+      // Validate partner_id exists in config
+      if (!config.partner_id || !config.partner_id[this.partnerId]) {
+        consola.error(`Template '${handle}' has no partner_id entry for partner ${this.partnerId}. Import the template to this partner first.`);
+        process.exit(1);
+      }
+
+      const templateId = config.partner_id[this.partnerId];
+      const templateContent = ReconciliationText.read(handle);
+
+      templates.push({
+        type: "Global::Partner::ReconciliationText",
+        id: String(templateId),
+        text: templateContent.text,
+        text_parts: templateContent.text_parts,
+      });
+    }
+
+    // Process account templates
+    for (const name of accountTemplates) {
+      const templateType = "accountTemplate";
+      const configPresent = fsUtils.configExists(templateType, name);
+
+      if (!configPresent) {
+        consola.error(`Config file for account template "${name}" not found`);
+        process.exit(1);
+      }
+
+      const config = fsUtils.readConfig(templateType, name);
+
+      // Validate partner_id exists in config
+      if (!config.partner_id || !config.partner_id[this.partnerId]) {
+        consola.error(`Template '${name}' has no partner_id entry for partner ${this.partnerId}. Import the template to this partner first.`);
+        process.exit(1);
+      }
+
+      const templateId = config.partner_id[this.partnerId];
+      const templateContent = AccountTemplate.read(name);
+
+      templates.push({
+        type: "Global::Partner::AccountDetailTemplate",
+        id: String(templateId),
+        text: templateContent.text,
+        text_parts: templateContent.text_parts,
+      });
+    }
+
+    // Process shared parts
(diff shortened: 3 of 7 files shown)


lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,246 @@
...
+    consola.info(`Starting sampler run with ${samplerParams.templates.length} template(s)...`);
+
+    // Start sampler run
+    const samplerResponse = await SF.createSamplerRun(partnerId, samplerParams);
+    const samplerId = samplerResponse.data.id || samplerResponse.data;
+
+    if (!samplerId) {
+      consola.error("Failed to start sampler run - no ID returned");
+      process.exit(1);
+    }
```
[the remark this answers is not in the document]
michieldegezelle: **Additional context:** This comment appears stale — the current version of the code already addresses this concern. Lines 48–55 now have an explicit null-safety guard:

```js
if (!samplerResponse?.data?.id) {
  consola.error("Failed to start sampler run - no ID returned");
  if (samplerResponse?.data) {
    consola.error(`Response data: ${JSON.stringify(samplerResponse.data)}`);
  }
  process.exit(1);
}
```

This can be resolved/dismissed.

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,246 @@
...
+  try {
+    consola.info(`Fetching status for sampler run ID: ${samplerId}`);
+
+    const response = await SF.readSamplerRun(partnerId, samplerId);
+
+    await handleSamplerResponse(response.data);
+  } catch (error) {
+    errorUtils.errorHandler(error);
+  }
+}
```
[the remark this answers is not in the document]
michieldegezelle: **Additional context:** This is already addressed in the current version. `checkStatus` now has an explicit guard before accessing `response.data`:

```js
if (!response || !response.data) {
  consola.error("Failed to fetch sampler run status. Is staging running?");
  process.exit(1);
}
```

Safe to resolve.

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,247 @@
...
+      }
+
+      // Build payload
+      const samplerParams = await this.#buildSamplerParams(templateHandles);
+
+      consola.info(`Starting sampler run with ${samplerParams.templates.length} template(s)...`);
+
+      // Start sampler run
+      const samplerResponse = await SF.createSamplerRun(this.partnerId, samplerParams);
+      const samplerId = samplerResponse.data.id || samplerResponse.data;
```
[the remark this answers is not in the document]
michieldegezelle: **Additional context:** The `data.id || data` fallback mentioned here no longer exists in the current code. After the `if (!samplerResponse?.data?.id)` guard, `samplerId` is now set simply as `samplerResponse.data.id`. Safe to resolve.

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,259 @@
...
+
+      const templateId = config.partner_id[this.partnerId];
+      const templateContent = AccountTemplate.read(name);
+
+      templates.push({
+        type: "Global::Partner::AccountDetailTemplate",
+        id: String(templateId),
+        text: templateContent.text,
+        text_parts: templateContent.text_parts,
+      });
```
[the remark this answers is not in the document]
michieldegezelle: **Additional context:** To be precise — `ReconciliationText.read` and `AccountTemplate.read` are both *synchronous* methods (not async/Promise-returning), so `await` is not the issue. The real concern is that both methods return `false` when `checkValidName` fails (e.g. `ReconciliationText.read` line 94: `if (!templateUtils.checkValidName(handle, this.TEMPLATE_TYPE)) return false`).

If `false` is returned, the code then does `templateContent.text` which throws `TypeError: Cannot read properties of false (reading 'text')`. The outer `try/catch` in `run()` will catch it via `errorUtils.errorHandler`, but the error message is confusing compared to the explicit `consola.error` + `process.exit` pattern used elsewhere in `#buildSamplerParams`. A guard like:

```js
const templateContent = ReconciliationText.read(handle);
if (!templateContent) {
  consola.error(`Reconciliation text "${handle}" could not be read`);
  process.exit(1);
}
```

...is the fix needed — not adding `await`.

lib/api/sfApi.js
```diff
@@ -717,6 +717,30 @@ async function getExportFileInstance(firmId, companyId, periodId, exportFileInst
...
+  const instance = AxiosFactory.createInstance("partner", partnerId);
+  try {
+    const response = await instance.post("liquid_sampler/run", attributes);
+    apiUtils.responseSuccessHandler(response);
+    return response;
+  } catch (error) {
+    const response = await apiUtils.responseErrorHandler(error);
+    return response;
+  }
+}
```
michieldegezelle, line 730: **💡 Suggestion:** The `samplerId` from the user-supplied `--id` flag is interpolated directly into the request path without any validation:

```js
instance.get(`liquid_sampler/${samplerId}`)
```

A value like `../partner_config` would produce the path `liquid_sampler/../partner_config`, which the server may normalize to a different endpoint. Since this comes from CLI user input (`options.id` in `bin/cli.js`), a simple guard in the CLI handler before calling `checkStatus` would close the hole:

```js
if (!/^\d+$/.test(options.id)) {
  consola.error("Invalid sampler ID: must be a numeric value");
  process.exit(1);
}
```

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,263 @@
...
+      // Validate at least one template specified
+      const { reconciliationTexts = [], accountTemplates = [], sharedParts = [] } = templateHandles;
+      if (reconciliationTexts.length === 0 && accountTemplates.length === 0 && sharedParts.length === 0) {
+        consola.error("You need to specify at least one template using -h, -at, or -s");
+        process.exit(1);
+      }
+
+      // Build payload
+      const samplerParams = await this.#buildSamplerParams(templateHandles, firmIds);
+
```
michieldegezelle, line 33: **💡 Suggestion:** The empty-handles guard inside `run()` (lines 36–40) duplicates the identical check already performed in `bin/cli.js` before `run()` is called. Because this path calls `process.exit(1)`, it makes `LiquidSamplerRunner.run()` hard to unit-test in isolation — a test that passes empty handles would kill the process rather than throw a catchable error.

Consider either removing the duplicate guard from `run()` (relying on the CLI layer) or replacing `process.exit(1)` here with a thrown `Error` so callers can handle it:

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,247 @@
...
+    switch (response.status) {
+      case "failed":
+        consola.error(`Sampler run failed: ${response.error_message || "Unknown error"}`);
+        break;
+
+      case "completed":
+        consola.success("Sampler run completed successfully");
+
+        if (response && response.result_url) {
+          await new UrlHandler(response.content_url).openFile();
```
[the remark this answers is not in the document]
michieldegezelle: **Additional context:** This comment appears to be based on a stale diff. In the current code, both the guard and the `UrlHandler` call consistently use `result_url`:

```js
if (response && response.result_url) {
  await new UrlHandler(response.result_url).openFile();
}
```

`content_url` is the field name used by the **export file instance** API (`exportFileInstanceGenerator.js`), which is a different endpoint. Unless the liquid sampler API is confirmed to return `content_url` instead of `result_url`, this is not a property mismatch — please verify against the sampler API contract before changing.

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,263 @@
...
+        consola.error(`Sampler run failed: ${response.error_message || "Unknown error"}`);
+        break;
+
+      case "completed":
+        consola.success("Sampler run completed successfully");
+
+        if (response && response.result_url) {
+          await new UrlHandler(response.result_url).openFile();
+        } else {
+          consola.warn("No URL returned");
```
michieldegezelle: **🟡 Minor:** A `completed` sampler run with no `result_url` only logs a warning and exits 0. For consistency with `ExportFileInstanceGenerator.#openUrl` (which uses `consola.error` when the download URL is missing), consider treating a missing URL as a failure so CI/scripts can detect it:

```js
} else {
  consola.error("Sampler completed but no result URL was returned");
  process.exit(1);
}
```

bin/cli.js
```diff
@@ -514,6 +515,44 @@ program
...
+  .option("-h, --handle <handles...>", "Specify reconciliation text handle(s) - can specify multiple")
+  .option("-at, --account-template <names...>", "Specify account detail template name(s) - can specify multiple")
+  .option("-s, --shared-part <names...>", "Specify shared part name(s) - can specify multiple")
+  .option("--firm-ids <firm-ids...>", "Specify firm ID(s) to run the sampler against - can specify multiple (optional)")
+  .option("--id <sampler-id>", "Specify an existing sampler ID to fetch results for (optional)")
+  .action(async (options) => {
+    // If an existing sampler ID is provided, fetch and display results
+    if (options.id) {
+      await new LiquidSamplerRunner(options.partner).checkStatus(options.id);
+      return;
```
michieldegezelle, line 543: **💡 Suggestion:** When `--id` is provided, template flags (`-h`, `-at`, `-s`) and `--firm-ids` are silently ignored because of the early return. Consider logging a warning if any of those are also present, to avoid users thinking they triggered a new run.

Toby-Masters-SF: Implemented a `conflicts` so cannot run both --id and other handles

bin/cli.js
```diff
@@ -486,6 +488,44 @@ program
...
+    const templateHandles = {
+      reconciliationTexts: handles,
+      accountTemplates: accountTemplates,
+      sharedParts: sharedParts,
+    };
+
+    const firmIds = options.firmIds || [];
+
+    await new LiquidSamplerRunner(options.partner).run(templateHandles, firmIds);
+  });
```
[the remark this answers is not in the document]
Toby-Masters-SF: Its a valid point, but I think its perhaps something we should implement as a method for either all CLI commands or none of them to enforce consistency.

Also less worried re this method as will always be called by the GH action

Something to add to the board for future changes? Could be handy to catch typos when running the CLI commands (e.g. accidentally typing in a String for a firm reference?)

lib/liquidSamplerRunner.js
Toby-Masters-SF: Removed the duplicate from `.run`

lib/liquidSamplerRunner.js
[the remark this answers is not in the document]
Toby-Masters-SF: Added in check for null/falsy template content

lib/api/sfApi.js
Toby-Masters-SF: Agreed and implemented

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,259 @@
...
+
+    return { templates, firm_ids: firmIds };
+  }
+
+  /**
+   * Poll for sampler run completion
+   * @param {number} partnerId - The partner ID
+   * @param {string} samplerId - The sampler run ID
+   * @returns {Promise<Object>} The completed sampler run
+   */
```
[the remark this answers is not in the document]
Toby-Masters-SF: Agreed & removed

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,259 @@
...
+  async #handleSamplerResponse(response) {
+    if (!response || !response.status) {
+      consola.error("Invalid sampler response");
+      process.exit(1);
+    }
+
+    switch (response.status) {
+      case "failed":
+        consola.error(`Sampler run failed: ${response.error_message || "Unknown error"}`);
+        break;
```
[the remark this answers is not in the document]
Toby-Masters-SF: Agree and updated to include break

lib/liquidSamplerRunner.js
Toby-Masters-SF: Agreed and updated. Should consider perhaps throwing the error though?

lib/cli/cliUpdater.js
```diff
@@ -5,6 +5,7 @@ const exec = promisify(require("child_process").exec);
 const { consola } = require("consola");
 const pkg = require("../../package.json");
 const { ChangelogReader } = require("./changelogReader");
+
 class CliUpdater {
   static #UPDATE_COMMAND = `sudo npm install -g ${pkg.repository.url}`;
```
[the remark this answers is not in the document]
Toby-Masters-SF: OOS for change and outdated

lib/exportFileInstanceGenerator.js
```diff
@@ -0,0 +1,103 @@
...
+          consola.error(`Export file generation failed or encountered an unexpected state. ${this.#details(exportFileInstanceId)}`);
+          return false;
+        }
+      }
+
+      return false;
+    } catch (error) {
+      errorUtils.errorHandler(error);
+    }
+  }
```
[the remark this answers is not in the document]
Toby-Masters-SF: OOS and outdated

lib/utils/urlHandler.js
```diff
@@ -0,0 +1,53 @@
...
+      const fileExtension = contentDisposition ? this.#identifyExtension(contentDisposition) : "";
+      const tempFilePath = path.resolve(require("os").tmpdir(), "silverfin", `${Date.now()}.${fileExtension}`);
+      fs.mkdirSync(path.dirname(tempFilePath), { recursive: true });
+      fs.writeFileSync(tempFilePath, response.data);
+
+      return tempFilePath;
+    } catch (error) {
+      errorUtils.errorHandler(error);
+    }
+  }
```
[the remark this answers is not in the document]
Toby-Masters-SF: OOS and outdated

lib/utils/urlHandler.js
```diff
@@ -0,0 +1,53 @@
...
+      const filePath = await this.#downloadFile();
+      await open(filePath);
+    } catch (error) {
+      consola.error(`Failed to open URL: ${this.url}`, error);
+    }
+  }
+
+  async #downloadFile() {
+    try {
+      const response = await axios.get(this.url, { responseType: "arraybuffer" });
```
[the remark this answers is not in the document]
Toby-Masters-SF: OOS and outdated

lib/utils/urlHandler.js
```diff
@@ -0,0 +1,53 @@
...
+      const contentDisposition = response.headers["content-disposition"];
+      const fileExtension = contentDisposition ? this.#identifyExtension(contentDisposition) : "";
+      const tempFilePath = path.resolve(require("os").tmpdir(), "silverfin", `${Date.now()}.${fileExtension}`);
+      fs.mkdirSync(path.dirname(tempFilePath), { recursive: true });
+      fs.writeFileSync(tempFilePath, response.data);
```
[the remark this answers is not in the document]
Toby-Masters-SF: OOS and outdated

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,260 @@
...
+        } else {  
+          consola.error("Sampler completed but no result URL was returned");  
+          process.exit(1);  
+          break; // eslint-disable-line no-unreachable
+        }
+        break;
+
+      case "pending":
+        
+      case "running":
```
Toby-Masters-SF, line 252: 🟠 **Major** — Empty `case "pending":` falls through to `running` without an explicit `break` or eslint directive, triggering `no-fallthrough` and failing the `test (22.x)` lint step in CI.

```javascript
      case "pending":
      // eslint-disable-next-line no-fallthrough
      case "running":
        consola.info(`Sampler run is still in progress. Current status: "${response.status}". Please check again later.`);
        break;
```

Alternatively combine into a single case block with `case "pending":` / `case "running":` using fall-through intentionally with the disable comment.

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,260 @@
...
+   */
+  async #fetchAndWaitSamplerResult(samplerId) {
+    let samplerRun = { status: "pending" };
+    const pollingDelay = 15000; // 15 seconds
+    const waitingLimit = 3600000; // 1 hour
+
+    spinner.spin("Running sampler...");
+    let waitingTime = 0;
+
+    while (samplerRun.status === "pending" || samplerRun.status === "running") {
```
Toby-Masters-SF: 🟠 **Major** — The polling loop calls `spinner.stop()` only on normal exit and timeout paths; if `SF.readSamplerRun` throws, the spinner is left running in the terminal.

```javascript
    try {
      while (samplerRun.status === "pending" || samplerRun.status === "running") {
        // ... polling ...
      }
      return samplerRun;
    } finally {
      spinner.stop();
    }
```

Toby-Masters-SF (approved):
Reviewed with changes made

lib/liquidSamplerRunner.js
Toby-Masters-SF: Implemented

lib/liquidSamplerRunner.js
Toby-Masters-SF: Added fix!

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,264 @@
...
+   * Resolve a template's partner-specific ID from its local config.
+   * Exits the process with a helpful message if the config is missing or has
+   * no partner_id entry for the current partner.
+   * @param {string} templateType - Config type ("reconciliationText" | "accountTemplate" | "sharedPart")
+   * @param {string} handle - Template handle/name
+   * @param {string} label - Human-readable label used in error messages
+   * @returns {string} The partner template ID
+   */
+  #resolveTemplateId(templateType, handle, label) {
+    if (!fsUtils.configExists(templateType, handle)) {
```
michieldegezelle, line 147: 🔴 **Security: Path Traversal via Template Handle**

The template handle parameter from `-h` flag is passed directly to `fsUtils.configExists()` and `fsUtils.readConfig()` without validation, allowing directory traversal.

**Risk:** User could run `silverfin run-sampler -p 123 -h '../../sensitive/data'` to read arbitrary config files outside the intended template directories.

**Fix:** Validate handle before use:
```javascript
if (!/^[a-zA-Z0-9_-]+$/.test(handle)) {
  consola.error(`Invalid template handle: ${handle}`);
  process.exit(1);
}
```

bin/cli.js
```diff
@@ -514,6 +515,55 @@ program
...
+      process.exit(1);
+    }
+
+    const templateHandles = {
+      reconciliationTexts: handles,
+      accountTemplates: accountTemplates,
+      sharedParts: sharedParts,
+    };
+
+    const firmIds = options.firmIds || [];
```
michieldegezelle, line 562: 🔴 **Bug: firmIds Type Mismatch**

The `--firm-ids` option receives strings from Commander.js, but the API expects numeric IDs. These strings are passed directly to the payload without conversion.

**Risk:** API receives `firm_ids: ['123', '456']` instead of `[123, 456]`, causing type validation failures or incorrect firm ID matching.

**Fix:** Convert strings to integers:
```javascript
const firmIds = (options.firmIds || []).map(id => parseInt(id, 10));
```

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,264 @@
...
+
+      consola.info(`Starting sampler run with ${samplerParams.templates.length} template(s)...`);
+
+      // Start sampler run
+      const samplerResponse = await SF.createSamplerRun(this.partnerId, samplerParams);
+
+      if (!samplerResponse?.data?.id) {
+        consola.error("Failed to start sampler run - no ID returned");
+        if (samplerResponse?.data) {
+          consola.error(`Response data: ${JSON.stringify(samplerResponse.data)}`);
```
michieldegezelle, line 42: 🟡 **Security: Credentials Exposed in Error Logs**

Line 42 logs the entire API response body via `JSON.stringify(samplerResponse.data)`, which could contain sensitive information like API keys, bearer tokens, or user data.

**Risk:** Secrets visible in terminal history, log files, and CI/CD logs.

**Fix:** Only log the error message without response details:
```javascript
if (samplerResponse?.data) {
  // Don't log full response - could contain secrets
  // consola.error(`Response data: ${JSON.stringify(samplerResponse.data)}`);
  // Just provide generic error message
}
```

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,264 @@
...
+   */
+  #resolveTemplateId(templateType, handle, label) {
+    if (!fsUtils.configExists(templateType, handle)) {
+      consola.error(`Config file for ${label} "${handle}" not found`);
+      process.exit(1);
+    }
+
+    const config = fsUtils.readConfig(templateType, handle);
+
+    if (!config.partner_id || !config.partner_id[this.partnerId]) {
```
michieldegezelle, line 154: 🔴 **Bug: Type Coercion in partner_id Lookup**

The `partnerId` is a string from the CLI option (e.g., `'123'`), but config file keys might be numeric (e.g., `123`). The lookup `config.partner_id[this.partnerId]` will fail to match.

**Scenario:** User runs `--partner 123`, CLI passes string `'123'`, but config has `{ partner_id: { 123: 'xyz' } }` with numeric key.

**Result:** Check fails incorrectly: `!config.partner_id['123']` doesn't find `123`.

**Fix:** Normalize to consistent type:
```javascript
if (!config.partner_id || !config.partner_id[String(this.partnerId)]) {
```

lib/api/sfApi.js
```diff
@@ -717,6 +717,30 @@ async function getExportFileInstance(firmId, companyId, periodId, exportFileInst
   }
 }
 
+async function createSamplerRun(partnerId, attributes) {
+  const instance = AxiosFactory.createInstance("partner", partnerId);
+  try {
+    const response = await instance.post("liquid_sampler/run", attributes);
+    apiUtils.responseSuccessHandler(response);
+    return response;
```
michieldegezelle, line 725: 🔴 **Architecture: API Response Pattern Inconsistency**

The new `createSamplerRun` and `readSamplerRun` return the full axios response object, while the existing pattern (`createExportFileInstance`, `getExportFileInstance`) returns only `response.data`.

**Risk:** This creates two conflicting patterns in the codebase. Client code must access `samplerResponse.data.id` instead of `samplerResponse.id`, violating the established architectural precedent.

**Fix:** Follow existing pattern by returning `response.data` instead of full response:
```javascript
// In sfApi.js createSamplerRun and readSamplerRun
return response.data;  // Not: return response;
```

michieldegezelle (commented):
## 📋 Comprehensive Review Summary

**Status:** 🛑 **REQUEST CHANGES** before merge

### Critical Issues (5 blockers)
✅ Posted inline comments above for:
1. **Path traversal** via template handle validation
2. **firmIds type mismatch** (strings vs numbers)
3. **Credentials exposure** in error logs
4. **partner_id type coercion** bug
5. **API response pattern inconsistency**

### Test Coverage Gap
⚠️ **Missing:** No unit tests for new `LiquidSamplerRunner` class or `run-sampler` CLI command. Other lib modules have corresponding `.test.js` files. Recommend:
- [ ] Add `tests/lib/liquidSamplerRunner.test.js`
- [ ] Add `tests/bin/cli/run-sampler.test.js`

### Additional Findings
🟡 **Medium:** Empty object validation (line 175), unvalidated URL from API  
🟢 **Low:** Integer overflow check, status pre-validation, code style (spacing + trailing whitespace)

### What's Good ✅
- All CodeRabbit findings properly addressed
- Error handling correctly structured
- CLI options and imports verified correct
- CI passing (565 tests)
- Version bump synchronized

### Estimated Fix Time
- High-severity fixes: **2-3 hours**
- Test coverage: **1-2 hours**
- **Total:** ~3-4 hours before re-review

**Please address blockers and request re-review.**

lib/liquidSamplerRunner.js
```diff
@@ -0,0 +1,264 @@
...
+
+        if (!samplerRun?.status) {
+          // process.exit() bypasses the finally block, so stop the spinner explicitly here.
+          spinner.stop();
+          consola.error("Failed to fetch sampler run status. Is staging running?");
+          process.exit(1);
+        }
+
+        waitingTime += pollingDelay;
+        // pollingDelay *= 1.05;
```
michieldegezelle: 💡 **Suggestion**

Leftover commented-out line with no accompanying config/flag to enable it. Consider removing it (or replacing with a real backoff implementation) to avoid dead code sitting in the polling loop.

```suggestion
        waitingTime += pollingDelay;
```

lib/liquidSamplerRunner.js
michieldegezelle: **Note:** Verified this is a pre-existing, codebase-wide pattern, not something introduced by this PR. `fsUtils.configExists`/`readConfig` build the path via `path.join(process.cwd(), FOLDERS[templateType], handle, "config.json")` for *every* template-handling call site in the codebase (`liquidTestRunner.js:225-233`, `templates/reconciliationText.js`, `templates/accountTemplate.js`, `templates/sharedPart.js`, `templates/exportFile.js`, `cli/stats.js`) — none of them validate the handle either. This is also a purely local CLI tool operated by the user against their own filesystem/credentials, so there's no privilege boundary being crossed (the "attacker" already has full local shell access). Given the systemic nature and lack of a real trust boundary, I don't think this PR should be the one to introduce handle validation in isolation — would suggest tracking as a follow-up applied consistently across all `fsUtils` callers if the team wants it.

bin/cli.js
michieldegezelle: **Note:** Same root cause as CodeRabbit's earlier comment on this exact line (thread starting at #discussion_r2741875054) — Toby-Masters-SF and CodeRabbit already discussed this and explicitly decided not to fix it now: this path is only invoked by the GH Action with controlled input, and a one-off numeric-coercion fix here would be inconsistent with the rest of the CLI (which doesn't coerce other numeric-looking args like `partnerId`/`firmId` elsewhere either). Tracked as a possible future follow-up (shared validation helper across CLI options) rather than a fix for this PR.

lib/liquidSamplerRunner.js
michieldegezelle: **Note:** `samplerResponse.data` here is the *response body* from the Silverfin `liquid_sampler/run` endpoint (the platform's own diagnostic payload for why the run failed to start), not the request payload/headers — it wouldn't contain API keys or bearer tokens (those live in the Authorization header, never echoed back in a response body). This also matches the existing repo-wide convention of logging full response bodies for debugging (see `apiUtils.js` `responseErrorHandler`, which already logs `JSON.stringify(error.response.data)` for 404/400/422 everywhere else in the codebase). Don't see evidence this specific log introduces a new credential-leak risk beyond what's already standard practice here.

lib/liquidSamplerRunner.js
michieldegezelle: **Note:** Verified this is incorrect — JSON object keys are always strings by spec, and JS bracket-notation property access always coerces the key to a string before lookup (`obj[123]` and `obj['123']` both resolve to the same property). There's no way for `JSON.parse` to produce an actual numeric key, so `config.partner_id[this.partnerId]` with a string `partnerId` will always match correctly regardless of how the JSON "looks". Confirmed empirically:
```
> JSON.parse('{"123":"abc"}')['123']
'abc'
> JSON.parse('{"123":"abc"}')[123]
'abc'
```

lib/api/sfApi.js
michieldegezelle: **Note:** Checked this against the actual codebase — the claimed "established precedent" is backwards. Of 84 functions in `sfApi.js`, 76 (including all the more recently added ones) `return response;` (full axios response) and only 8 (an older minority, e.g. `createExportFileInstance`/`getExportFileInstance`) `return response.data;` directly. `createSamplerRun`/`readSamplerRun` follow the majority convention, and they need to return the full response because `liquidSamplerRunner.js` relies on `samplerResponse?.data?.id` / `response?.data?.status` optional-chaining for null-safety on the whole response, not just on `.data`. Changing to `return response.data` would actually break that null-safety pattern (since `response.data` itself could be undefined on error, whereas the current code guards on the full response object).

lib/liquidSamplerRunner.js
michieldegezelle: **Note:** Verified this doesn't hold — ESLint's `no-fallthrough` rule explicitly ignores cases with an empty body falling through to the next `case`; it only flags fallthrough when a non-empty case block is missing a `break`/`return`/`throw`. Ran `npx eslint .` (and `npx eslint lib/liquidSamplerRunner.js` directly) on the current HEAD — zero errors/warnings. Also, `lint` and `test` are separate `package.json` scripts, and CI's `test (22.x)` check only runs `jest`, not `eslint` — so this wouldn't fail that check either way. No action needed.

lib/liquidSamplerRunner.js
michieldegezelle: **Stale:** Verified against current code — the whole polling loop is now wrapped in `try { ... } finally { spinner.stop(); }` (lines 196-226), so `spinner.stop()` runs on *any* exit from the try block, including a thrown error from `SF.readSamplerRun`. Only the `process.exit()` paths inside the loop need (and have) an explicit `spinner.stop()` call first, since `process.exit()` bypasses `finally` — which is exactly what the inline comments there say. This is fixed. No action needed.

lib/liquidSamplerRunner.js
michieldegezelle: **False positive.** `samplerResponse.data` here is the JSON body of `POST liquid_sampler/run` — the sampler run object (`id`, `status`, etc., per its usage as `samplerResponse.data.id` right above). It doesn't carry auth tokens or secrets; those are attached via axios request interceptors and never appear in response bodies. There's no concrete evidence this specific endpoint returns sensitive data, so resolving as speculative/not applicable.

lib/liquidSamplerRunner.js
michieldegezelle: **False positive.** JS object/bracket-notation key access always coerces to string, regardless of whether the key was originally numeric or string (and JSON keys are always strings after parsing anyway). Verified empirically:

```js
const config = JSON.parse(JSON.stringify({ partner_id: { 123: 'xyz' } }));
config.partner_id['123'] // 'xyz'
config.partner_id[123]   // 'xyz'
```

Both lookups resolve identically, so `config.partner_id[this.partnerId]` works whether `this.partnerId` is a string or number. Resolving.

lib/liquidSamplerRunner.js
Toby-Masters-SF: As discussed OOS

bin/cli.js
Toby-Masters-SF: OOS

lib/api/sfApi.js
Toby-Masters-SF: OOS

lib/liquidSamplerRunner.js
Toby-Masters-SF: Agreed to delete

PR #247, opened 2026-02-15
Add workflow option to stats command

### LINKED ISSUE #241: Workflow test statistics
Update the `stats` CLI command to generate statistics on a workflow level.

Workflows for each Market should be stored in a new "Workflows" folder. Each individual workflow folder with be a single JSON file in the following format, which will be read by the CLI:

```
{
  "workflow_handle": "uk_corporation_tax"
  "name_nl": "UK Corporation Tax"
  "templates": [
    "ct_contents",
    "ct_returns_information"
    ...
  ]  
}
```

Within the CLI a new `-w/--workflow` option would be added to the `stats` command. If no workflow is provided, a test coverage statistics for _all_ workflows will be provided. If a workflow is provided only the test coverage statistics for that specified workflow will appear.

As part of the above work, a workflow model should be created (similar to the reconciliation or shared part models)

Finally, to setup the above JSON files an AI flow should be crafted. Currently have a 2-step plan:

1.  Use a screenshot of a workflow with Gemeni to produce a text list of reconciliation names
2. Provide the text list to Cursor to then look up the handles with that name and use them to populate a JSON workflow file.

Considered a single JSON file with all workflows but would be much more difficult to read and may be more difficult to adapt to future changes in the CLI

**Additional context**
When addressing problem, bear in mind this could be the first step toward incorporating workflows and workflow working screens into the CLI.

Toby-Masters-SF:

## Description

## Testing Instructions

Steps:

1. ...
2. ...
3. ...

## Author Checklist

## Reviewer Checklist

### Plan
- [ ] Create a new `getWorkflowTemplateSummary(workflowHandle)` function
- [ ] Create a new `listWorkflowTemplates` function which can take different classes?

### Testing notes
- Reconciliations (cross-referenced using AI to interrogate). Reason for difference is multiple `yml` files in the same `tests` folder so may need to update the Regex expression: `.*${FOLDER}/${templatePattern}/tests/.*_liquid_test.*.y(a)?ml`. Resolved by temporarily deleting "extra" tests in test branch
  - Total - 58: CORRECT
  - Externally maanged - 58: CORRECT
  - YAML - at least 1 test - 49: TO CHECK? (AI GIVING 46)
  -  YAML - 2+ tests - 45: TO CHECK? (AI GIVING 43)
  

### Quick Notes
 - Plan is to create a series of _parallel_ functions to avoid overcomplicating the existing functions 
   - May also wish to consider moving to a separate file in the `cli` folder? 
 - Review function `getTemplatesSummary` --> Can it be co-opted for the purposes of this workflow? Take the "Reconciliations" section as an example!
    - [TBC]

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -570,6 +570,7 @@ program
   .command("stats")
   .description("Generate an overview with some statistics")
   .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
+  .option("-w, --workflow [handle]", "Optionally filter test statistics by workflow (workflow handle is optional)")
   .action((options) => {
     stats.generateOverview(options.since);
   });
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -572,7 +572,12 @@ program
   .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
   .option("-w, --workflow [handle], Optionally filter test statistics by workflow (optional)")
   .action((options) => {
-    stats.generateOverview(options.since, options.workflow);
+    if (options.workflow) {
+      const workflowHandle = typeof options.workflow === "string" ? options.workflow : undefined;
+      stats.generateWorkflowOverview(options.since, workflowHandle);
+    } else {
+      stats.generateOverview(options.since);
+    }
   });
 
 // Set/Get FIRM ID
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -570,9 +570,9 @@ program
   .command("stats")
   .description("Generate an overview with some statistics")
   .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
-  .option("-w, --workflow [handle]", "Optionally filter test statistics by workflow (workflow handle is optional)")
+  .option("-w, --workflow [handle], Optionally filter test statistics by workflow (optional)")
   .action((options) => {
-    stats.generateOverview(options.since);
+    stats.generateOverview(options.since, options.workflow);
   });
 
 // Set/Get FIRM ID
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -626,6 +626,7 @@ program
   .command("stats")
   .description("Generate an overview with some statistics")
   .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
+  .option("-w, --workflow [handle]", "Optionally filter test statistics by workflow (workflow handle is optional)")
   .action((options) => {
     stats.generateOverview(options.since);
   });
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -626,9 +626,9 @@ program
   .command("stats")
   .description("Generate an overview with some statistics")
   .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
-  .option("-w, --workflow [handle]", "Optionally filter test statistics by workflow (workflow handle is optional)")
+  .option("-w, --workflow [handle], Optionally filter test statistics by workflow (optional)")
   .action((options) => {
-    stats.generateOverview(options.since);
+    stats.generateOverview(options.since, options.workflow);
   });
 
 // Set/Get FIRM ID
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -626,22 +626,36 @@ program
   .command("stats")
   .description("Generate an overview with some statistics")
   .requiredOption("-s, --since <date>", "Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
-  .option("-w, --workflow [handle]", "Filter the statistics by workflow. Without a handle, every workflow stored in the workflows folder is used (optional)")
+  .option(
+    "-w, --workflow [handle]",
+    "Filter the statistics by workflow. Pass a handle (--workflow <handle>) to report on a single workflow, or use --workflow on its own to report on every workflow stored in the workflows folder (optional)"
+  )
   .action(async (options) => {
+    // Check if the since date is in the correct format
     cliUtils.checkDateFormat(options.since);
-    // Commander sets workflow to true when the flag is used without a value
-    if (typeof options.workflow === "undefined") {
+    // Commander gives three distinct values: undefined when the workflow flag is absent,
+    // true when it is used without a value, and the text when a handle is given
+    if (options.workflow === undefined) {
       await stats.generateOverview(options.since);
       return;
     }
-    // A blank handle (e.g. an unset --workflow "$HANDLE") is falsy, so it would otherwise be read as "no handle given" and include every workflow
-    if (typeof options.workflow === "string" && options.workflow.trim() === "") {
-      consola.error(`An empty workflow handle was provided. Please pass a handle (--workflow <handle>) or use --workflow on its own to include every workflow`);
+    // The workflow flag was used on its own, so no handle is passed on and the statistics
+    // cover every workflow stored in the workflows folder
+    let workflowHandle;
+    if (options.workflow !== true) {
+      // A handle was typed. Surrounding spaces are a quoting accident, and a blank handle must
+      // never be read as "no handle given", so it is checked before it reaches the statistics
+      workflowHandle = options.workflow.trim();
+      // The same command without a handle reports on every workflow stored, which is the
+      // quickest way for the user to see which handles exist
+      cliUtils.checkHandleFormat(workflowHandle, "workflow handle", `silverfin stats --since ${options.since} --workflow`);
+    }
+    const reported = await stats.generateWorkflowOverview(options.since, workflowHandle);
+    // Not a single workflow could be reported on. This is the end of the run, so it is
+    // where the command stops
+    if (!reported) {
       process.exit(1);
     }
-    // When a handle was typed we pass it on without any surrounding spaces, otherwise the flag was used on its own (Commander gives true instead of text) and undefined means "every workflow"
-    const workflowHandle = typeof options.workflow === "string" ? options.workflow.trim() : undefined;
-    await stats.generateWorkflowOverview(options.since, workflowHandle);
   });
 
 // Set/Get FIRM ID
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -628,7 +628,12 @@ program
   .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
   .option("-w, --workflow [handle], Optionally filter test statistics by workflow (optional)")
   .action((options) => {
-    stats.generateOverview(options.since, options.workflow);
+    if (options.workflow) {
+      const workflowHandle = typeof options.workflow === "string" ? options.workflow : undefined;
+      stats.generateWorkflowOverview(options.since, workflowHandle);
+    } else {
+      stats.generateOverview(options.since);
+    }
   });
 
 // Set/Get FIRM ID
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -625,15 +625,23 @@ program
 program
   .command("stats")
   .description("Generate an overview with some statistics")
-  .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
-  .option("-w, --workflow [handle], Optionally filter test statistics by workflow (optional)")
-  .action((options) => {
-    if (options.workflow) {
-      const workflowHandle = typeof options.workflow === "string" ? options.workflow : undefined;
-      stats.generateWorkflowOverview(options.since, workflowHandle);
-    } else {
-      stats.generateOverview(options.since);
+  .requiredOption("-s, --since <date>", "Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
+  .option("-w, --workflow [handle]", "Filter the statistics by workflow. Without a handle, every workflow stored in the workflows folder is used (optional)")
+  .action(async (options) => {
+    cliUtils.checkDateFormat(options.since);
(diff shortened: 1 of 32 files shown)


bin/cli.js
```diff
@@ -570,8 +570,14 @@ program
   .command("stats")
   .description("Generate an overview with some statistics")
   .requiredOption("-s, --since <date>, Specify the date which is going to be used to filter the data from (format: YYYY-MM-DD) (mandatory)")
+  .option("-w, --workflow [handle], Optionally filter test statistics by workflow (optional)")
   .action((options) => {
-    stats.generateOverview(options.since);
+    if (options.workflow) {
+      const workflowHandle = typeof options.workflow === "string" ? options.workflow : undefined;
```
michieldegezelle: **`--workflow` without a value silently runs all-workflows scan instead of erroring.**

The option is declared with optional-argument syntax `[handle]`, so running `silverfin stats -s 2024-01-01 --workflow` (flag present, no value) causes Commander to set `options.workflow = true` (boolean). The `if (options.workflow)` check on the line above is truthy, but this `typeof === 'string'` guard then sets `workflowHandle = undefined`. As a result, `generateWorkflowOverview(sinceDate, undefined)` is called and silently reads every file in the `workflows/` folder — the user gets a full-repo scan with no indication their argument was dropped.

Consider using `<handle>` (required argument) instead of `[handle]`, or explicitly check `options.workflow === true` and print an error.

lib/cli/stats.js
```diff
@@ -400,4 +597,53 @@ function saveOverviewToFile(row) {
...
+    "Account Templates - yaml files with at least two tests (%)",
+  ];
+  const ROW_HEADER = `${COLUMNS.join(";")}`;
+  const CSV_PATH = `./stats/${workflowHandle}_stats.csv`;
+  // Create file and header columns
+  if (!fs.existsSync("./stats")) {
+    fs.mkdirSync("stats");
+  }
+  if (!fs.existsSync(CSV_PATH)) {
+    fs.writeFileSync(CSV_PATH, ROW_HEADER, (err) => {
```
michieldegezelle, line 641: **`fs.writeFileSync` silently ignores the callback passed as the third argument.**

`writeFileSync` is synchronous and its third parameter is an `options` object (encoding, mode, flag) — not an error handler. The callback `(err) => { consola.error(err); }` is coerced to an options object and silently discarded. Any write failure (e.g. permission denied, disk full) throws an uncaught synchronous exception rather than being handled. The same issue exists at line 592 in `saveOverviewToFile`.

Remove the callback and either let the exception propagate naturally, or wrap the call in a `try/catch`.

lib/cli/stats.js
```diff
@@ -10,16 +10,33 @@ async function generateOverview(sinceDate) {
...
 
+async function generateWorkflowOverview(sinceDate, workflowHandle) {
+  const TODAY = new Date().toJSON().toString().slice(0, 10);
+  let workflowHandles;
+  if (workflowHandle) {
+    workflowHandles = [workflowHandle];
+  } else {
+    const workflowsFolder = path.join(process.cwd(), "workflows");
+    workflowHandles = fs.readdirSync(workflowsFolder).map((file) => path.basename(file, ".json"));
+  }
```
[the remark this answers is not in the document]
michieldegezelle: Additional context: the same `readdirSync` line also lacks a filter for non-JSON files. Every entry in the folder is mapped through `path.basename(file, '.json')` — so `.DS_Store`, `README.md`, a subdirectory, etc. produce invalid handles (e.g. `.DS_Store` unchanged). `getWorkflow('.DS_Store')` then looks for `workflows/.DS_Store.json`, doesn't find it, and calls `process.exit(1)`, aborting the entire run before any remaining valid workflows are processed. The fix proposed above should also include a `.filter(f => f.endsWith('.json'))` before the `.map(...)`.

lib/cli/stats.js
```diff
@@ -35,7 +52,9 @@ async function yamlFilesActivity(sinceDate) {
   }
 
   // Files to Search (YAML)
-  const YAML_EXPRESSION = `.*/.*/tests/.*_liquid_test.*.y(a)?ml`;
+  const templatesInWorkflow = workflow ? workflow.templates.reconciliations.concat(workflow.templates.accounts) : [];
+  const TEMPLATE_PATTERN = workflow ? `(${templatesInWorkflow.join("|")})` : `.*`;
+  const YAML_EXPRESSION = `.*/${TEMPLATE_PATTERN}/tests/.*_liquid_test.*.y(a)?ml`;
   const fileTypeRegExp = RegExp(YAML_EXPRESSION, "g");
```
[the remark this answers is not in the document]
michieldegezelle: The identical pattern exists in `countYamlFiles` (line 99), where `TEMPLATE_PATTERN` is built the same way from `templatesInWorkflow`. The `escapeRegex` helper proposed above needs to be applied there too, otherwise workflow-filtered YAML counts will have the same unescaped-metacharacter exposure as the activity scan.

michieldegezelle (commented):
A few additional code-review notes not covered by existing threads.

Can you also update the changelog, bump the version in package.json and do a npm update to update the package-lock? Thanks

lib/cli/stats.js
```diff
@@ -35,7 +52,9 @@ async function yamlFilesActivity(sinceDate) {
   }
 
   // Files to Search (YAML)
-  const YAML_EXPRESSION = `.*/.*/tests/.*_liquid_test.*.y(a)?ml`;
+  const templatesInWorkflow = workflow ? workflow.templates.reconciliations.concat(workflow.templates.accounts) : [];
+  const TEMPLATE_PATTERN = workflow ? `(${templatesInWorkflow.join("|")})` : `.*`;
```
michieldegezelle, line 56: **Empty template list produces a broken regex pattern.**

When a workflow has no reconciliations and no accounts, `templatesInWorkflow` is `[]`, so `TEMPLATE_PATTERN` becomes `()` and `YAML_EXPRESSION` effectively matches all YAML test paths — the opposite of filtering. Consider returning early (or skipping YAML activity) when the combined list is empty.

lib/cli/stats.js
```diff
@@ -74,10 +93,12 @@ async function yamlFilesActivity(sinceDate) {
 // Count how many YAML files are stored. We base on the presence of a non empty file
 // Count how many unit tests are stored. We base on the presence of a title for each unit test
 // Count how many YAML files have at least two unit tests
-async function countYamlFiles(templateType) {
+async function countYamlFiles(templateType, templatesInWorkflow) {
   const files = fsUtils.listExistingFiles("yml");
   const FOLDER = fsUtils.FOLDERS[templateType];
-  const YAML_EXPRESSION = `.*${FOLDER}/.*/tests/.*_liquid_test.*.y(a)?ml`;
+  const TEMPLATE_PATTERN = templatesInWorkflow ? `(${templatesInWorkflow.join("|")})` : `.*`;
+  // Issue with counting multiple yml files in the same tests folder?
```
michieldegezelle, line 100: Leftover open question in a comment (`Issue with counting multiple yml files in the same tests folder?`). Either confirm this is not a problem and remove the comment, or document/fix the edge case.

lib/cli/stats.js
```diff
@@ -255,8 +276,99 @@ async function getTemplatesSummary() {
...
+      yamlFiles: 0,
+      yamlFilesPerc: 0,
+      unitTests: 0,
+      yamlFilesWithAtLeastTwoTests: 0,
+      yamlFilesWithAtLeastTwoTestsPerc: 0,
+    },
+  };
+
+  // Fetch workflow
+  // Assume no empty items if present within the workflow.
```
michieldegezelle, line 320: **Global vs workflow stats count different populations.**

`generateOverview` counts only *non-empty* templates (`listNonEmptyTemplates`), while workflow mode counts every handle listed in the workflow JSON. That is probably fine, but worth documenting in the CLI help or PR description so users do not compare the two totals directly.

lib/cli/stats.js
```diff
@@ -255,8 +276,99 @@ async function getTemplatesSummary() {
...
+  summary.all.yamlFiles = summary.reconciliations.yamlFiles + summary.accountTemplates.yamlFiles;
+  summary.all.yamlFilesPerc = percentageRoundTwo(summary.all.yamlFiles, summary.reconciliations.total + summary.accountTemplates.total);
+  summary.all.unitTests = summary.reconciliations.unitTests + summary.accountTemplates.unitTests;
+  summary.all.yamlFilesWithAtLeastTwoTests = summary.reconciliations.yamlFilesWithAtLeastTwoTests + summary.accountTemplates.yamlFilesWithAtLeastTwoTests;
+  summary.all.yamlFilesWithAtLeastTwoTestsPerc = percentageRoundTwo(summary.all.yamlFilesWithAtLeastTwoTests, summary.reconciliations.total + summary.accountTemplates.total);
+
+  return summary;
+}
+
+async function getYamlSummary(sinceDate, templatesInWorkflow) {
```
michieldegezelle, line 370: Minor: the second parameter is named `templatesInWorkflow` but callers pass the full `workflow` object into `yamlFilesActivity`. Renaming to `workflow` would match actual usage.

bin/cli.js
Toby-Masters-SF: This is part of the explicit design. We want to a general stats --workflow run and for a user to pull stats for a _specific_ workflow if provided

lib/cli/stats.js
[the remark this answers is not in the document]
Toby-Masters-SF: Workflow discovery now lives in fsUtils.getAllWorkflowHandles, which returns [] when the folder does not exist, and reportOnAllWorkflows reports that through errorUtils.noWorkflowsStored() rather than exiting inside the read.

lib/cli/stats.js
Toby-Masters-SF: Fixed. buildTemplatePattern returns undefined for an empty list instead of an empty alternation, and both callers bail out early — yamlFilesActivity explains no YAML changes can be counted and skips the git scan, countYamlFiles returns zeroes.

lib/cli/stats.js
Toby-Masters-SF: Renamed — the signature is now `yamlFilesActivity(sinceDate, workflow)`.

lib/cli/stats.js
Toby-Masters-SF: Right on both counts — callback was coerced into an options object and discarded, and saveOverviewToFile had it too.

- Both call sites now go through one private helper, `writeStatisticsRow(csvPath, rowHeader, row)` — they differed only in path + header
try/catch around folder creation, header write and append
- Raw error to `consola.debug` so `-v` still shows it
- Reported via new `errorUtils.statisticsNotWritten` — names the CSV path, adds "(no permission to write to it)" for EACCES
- Catch, no exit: summary is already on screen by then, so the run finishes normally instead of a stack trace for a file left open in Excel
- `reportOnWorkflow` ignores the returned boolean on purpose — a failed write isn't a skipped workflow in the batch summary, stats were gathered, only the CSV row lost
- Test makes `appendFileSync` throw `EACCES`, asserts path + reason reported, no exit. Also checked by hand: `chmod 444 stats/overview.csv` → `The statistics could not be written to ./stats/overview.csv (no permission to write to it). They are shown above`

Please let me know if you'd rather it propagated to a non-zero exit so CI fails on a lost row.

lib/cli/stats.js
Toby-Masters-SF: Good catch — it was a real problem, so I've fixed it rather than removing the comment.

The bug: the counts were per file, but every one of them is reported as a percentage of a template count. A template with two *_liquid_test.yml files in its tests/ folder was counted twice, so yaml files (%) could exceed 100%. yaml files with at least two tests had the same flaw in reverse: two files with one unit test each meant the template had two unit tests, but neither file qualified on its own, so it wasn't counted at all.

The fix (countYamlFiles): unit tests are now accumulated into a testsPerTemplate map keyed by the template folder name, taken from a capture group in the path regex. A template is counted once no matter how many test files it holds, and its unit tests are summed across them. Related changes that fell out of it:

- `buildTemplatePattern` returns `([^/]+)` instead of `.*` for the no-filter case — `.*` is greedy and would swallow the rest of the path, so it can't identify the template. The group is now always capturing.
- Dropped the g flag on the regex: we want the first match and its capture group, not a list.
- Renamed throughout: yamlFiles → templatesWithTests, yamlFilesWithAtLeastTwoTests → templatesWithAtLeastTwoTests, plus the matching console labels ("Templates with YAML tests") and CSV headers ("All - templates with yaml tests" etc.). The old names were the reason the file/template mismatch went unnoticed.
- The YAML parse catch no longer swallows silently — it logs at debug level. An unparseable file still marks the template as having tests; the template keeps the unit tests found in its other files.

The header rename changes the column labels in stats/overview.csv and stats/<workflow>_stats.csv. Column order and count are unchanged, so existing sheets keep working, but the values in the "templates with yaml tests" columns will now be lower where a template had several test files.

Coverage added in `tests/lib/cli/stats.test.js` — a new a template with more than one liquid test file block: template counted once with unit tests summed (repository overview and workflow overview), a template qualifying for "at least two tests" only on its combined total, and a regression case asserting the percentage never exceeds 100%.

lib/cli/stats.js
Toby-Masters-SF: Documented, and it turned out to be hiding a bug.

**Documented** (b2713dd):
- README has the same as a two-bullet comparison, plus the note that a template counts once however many liquid test files it holds
- CHANGELOG entry for 1.59.0 covers both

**The bug** (9382d21) — chasing the population difference turned up why counting every listed handle was worse than just inconsistent:
- For a handle listed in the workflow but not in the repo, `listExternallyManagedTemplates` → `fsUtils.readConfig` → `createConfigIfMissing` **created** the template folder and a `config.json` with `externally_managed: true`
- So a typo in a workflow file scaffolded a template through a read-only reporting command, counted it in the totals, and counted it as externally managed off the back of that generated default
- Now `reportOnWorkflow` narrows the workflow to what the repo holds (via `fsUtils.getAllTemplatesOfAType`) before any config is read, so nothing is created. Missing handles are named in one warning per workflow
- Filtering happens once at the top rather than inside `getWorkflowTemplateSummary`, so the counts and the `git whatchanged` scan see the same population

Verified: workflow listing `rec_typo` → warning names it, excluded from totals, `git status` shows no new folder. Four tests cover it, three of which fail without the change.

robindeclercq (approved):

Issue #240, opened 2026-01-19
[Feature Request] - Allow folder structure for a reconciliation template


Benjvandam:
**Description of the issue**
Platform clients showed that they created structure in the parts of their templates by creating folders. This would be interesting to try out for more complicated templates (such as immovable income in Belgium PIT package). This folder structure would potentially break the toolchain CLI. 

**Describe the solution you'd like**
Find out if and why the toolchain CLI would not work with subfolders within a template folder, and refactor it so that it would work.

**Describe alternatives you've considered**
I don't really see an alternative except for not implementing this folder structure

robindeclercq:
Also see: https://github.com/silverfin/silverfin-cli/issues/136

PR #248, opened 2026-02-16
Allow folder structure in config text parts

### LINKED ISSUE #136: Parts in subfolders removed from config automatically
**Describe the bug**
I'm encountering an issue in VSCode where the config keeps automatically adjusting itself. It removes text_parts. I made a video to illustrate: https://www.loom.com/share/c99faf521d3b4094a4bebeca124bd7ee?sid=ea801996-d0ba-4450-a555-90ffad21bca8
The way it shows when I discard changes is what is saved on the main branch. Each time I discard them, a few seconds later the changes show up again. This means I can't visualize my code in a demo firm, because the parts I am working in keep getting removed from the config.

The parts that get removed are saved in subfolders :
liquid-toolchain__Channel__-_Silverfin_-_Slack

The parts in blue are retained, while the parts in red are discarded.

**To Reproduce**
1. Go to the config file of the template with parts in subfolders. The parts are automatically removed already.
2. Click on 'discard changes'. This puts the parts back
3. Wait a few seconds
4. The parts get removed again automatically

**Expected behavior**
Parts in subfolders should remain in the config

See description above

 - OS: iOS
 - CLI version 1.26.11

**Additional context**
There is a workaround: disable the Silverfin extension in the CLI, discard the changes to the config, commit, and then enable the extension again. The issue doesn't reoccur after that for a while.

Benjvandam:
Fixes # https://github.com/silverfin/silverfin-cli/issues/240

## Description

This feature allows you to create a folder structure within the text_parts of a reconciliation and account template. First it builds a map of the actual folder structure. Then based on the map, the text parts in the config are updated to to reflect the correct path. If a file was not mentioned in the config file, it will not be suddenly be included. 

## Testing Instructions

Steps:

1. Update the folder structure in your files. Add folders and subfolders and shift your files text_parts around
2. Update/ push your files to the platform or run tests
3. The config file will automatically be updated to reflect the correct path and everything else keeps working

Keep into account that you may have to disable the extension for testing. Because it seems that when you move files around in subfolders, the Extension automatically removes them from the config when you save. In order for it to work properly, we may need to update the Extension as well still.

## Author Checklist

## Reviewer Checklist


michieldegezelle (changes requested):
Changes look good, but I discussed this with Agustin and we would slightly change the scope of this ticket (due to the stability of the auto-update). 

Could you make sure that all function calls correctly look at subfolders (which is already the case). The auto update of the config file should be skipped for now

1) It is not always so stable (e.g. when deleting a file)
2) It only works when a command is done (it is static), so ideally this is something for a future update in the extension.

This means that we should rely on the user to update the config file manually for now (but this is acceptable). 

In parallel, we should then prepare an update in the extension to:
- Make sure that subfolders are allowed
- removeDeletedParts() (templateUtils.ts) still works for subfolders
- “Open file” link (panelTemplateParts.ts) will correctly go to the file, even if it is in the subfolder


In a later stage, we can then look at the extension to automatically update the config as soon as the user saves, but let's keep that for later.

Benjvandam:
I’ve removed the auto-update behaviour for the config files and kept the utility that maps the actual file structure. Subfolders are still supported: they’re not a problem for the CLI, but the API only receives template files and no folder structure, so the Silverfin platform doesn’t preserve it. When you re-import a template that was already imported before, you get the flat file structure back from the API, so in that case you’ll need to re-create the folder structure (and update the config) manually.

michieldegezelle (approved):
@Benjvandam Could you just remove the temp folder from this PR? Approving already though. Let's wait until the extension works again before merging

Issue #242, opened 2026-01-21
[Feature Request] - Allow to push test data to company


Benjvandam:
**Is your feature request related to a problem? Please describe.**
The idea would be to allow a test to be pushed to a particular company. That would allow users to not have to manually re-create data in the company of the live environment when there is a bug for example. 

**Describe the solution you'd like**
Through the CLI, a command that includes the test and a firm and company that this test should be pushed to.

**Describe alternatives you've considered**
Current way of working using staging, or allow for stagings to be spun up a lot easier. But this approach would be a lot more lightweight.

**Additional context**
This was already "vibe coded". So we already have a good idea of what this could look like. It is just a matter of implementing this properly into the existing CLI

BenjaminLangenakenSF:
PR #260 implements this. Its `update-text-properties` command bulk-seeds a company from a Liquid Test YAML — exactly the ask here ("a command that includes the test and a firm and company that this test should be pushed to").

Two caveats before treating it as the fix:

1. #260 is much broader than this issue — 11 new commands, 14 new `lib/` modules, a dependency on `silverfin-ls`. I've suggested there that the three commands relevant to this issue (`update-text-properties`, `get-results`, `describe-inputs`) split out from the static-analysis half.
2. @Benjvandam has left, so #260 needs a new owner.

Deliberately not adding a `Closes #242` link yet — that would auto-close this on merge, which is only correct if the narrowed split happens.

BenjaminLangenakenSF:
Correcting my comment above: `update-text-properties` has **already merged to main**, in `Implement reverse test (#249)` — not pending in #260 as I said.

[`bin/cli.js#L590-L600`](https://github.com/silverfin/silverfin-cli/blob/e6b5645737989efb73eaa41f5a74cd28c73433f3/bin/cli.js#L590-L600) — takes a company URL plus a test name, reads the test from the template's `tests/` folder and uploads it, with `--dry-run` and `-y` to skip the prompt. That is this issue's ask: *"a command that includes the test and a firm and company that this test should be pushed to."*

#260 extends the same command (adds `--file` handling and the wider live-bridge surface) but is not what delivers it.

Looks closeable. Worth someone running it against a real company first to confirm it behaves end to end — I've only verified the command exists on main and reads the right inputs.

PR #249, opened 2026-03-16
Implement reverse test

Benjvandam:
Fixes # https://github.com/silverfin/silverfin-cli/issues/242

## Description

New `update-text-properties` command that uploads custom text properties from a Liquid Test YAML file to a reconciliation in a company file. It extracts custom data from all 4 levels: company, period, reconciliation, and account.

Supports `--handle` for faster YAML file lookup and `--dry-run` to preview the payload without uploading.

## Testing Instructions

Steps:

1. Navigate to a market repo (e.g. `be_market`) that has reconciliation templates with liquid test YAML files
2. Open a reconciliation in a company file on Silverfin and copy the full URL
3. Run with dry-run to verify the payload:
   ```
   silverfin update-text-properties -u "<url>" -t <test_name> -h <handle> --dry-run
   ```
4. Verify the output shows the correct properties extracted from the YAML test
5. Run without dry-run to push the properties:
   ```
   silverfin update-text-properties -u "<url>" -t <test_name> -h <handle>
   ```
6. Verify in the Silverfin UI that the custom properties are updated on the reconciliation
7. Test with a YAML file that has company-level custom data (under `data.company.custom`) and verify those are also pushed
8. Test without `--handle` to verify it scans all template folders and finds the test automatically

## Author Checklist

## Reviewer Checklist


bin/cli.js
```diff
@@ -527,6 +529,120 @@ program
...
+  .command("update-text-properties")
+  .description("Upload custom text properties from a Liquid Test YAML file to a reconciliation in a company file")
+  .requiredOption("-u, --url <url>", "Specify the full Silverfin URL of the reconciliation in the company file (mandatory)")
+  .requiredOption("-t, --test <test-name>", "Specify the name of the test to use as data source (mandatory)")
+  .option("-h, --handle <handle>", "Specify the reconciliation handle to narrow down the YAML file search (optional)")
+  .option("--dry-run", "Only transform and display the properties without uploading (optional)", false)
+  .action(async (options) => {
+    // Parse URL to extract IDs
+    const urlData = liquidTestUtils.extractURL(options.url);
+    const { firmId, companyId } = urlData;
```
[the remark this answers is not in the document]
Benjvandam: The purpose of the command is to set up a full test scenario in a company file — that requires pushing custom data to all levels and templates referenced in the test, not just the single reconciliation from the URL. The URL is used to identify the target company/firm/period, but the test YAML defines the complete data set needed for that scenario to work.

bin/cli.js
```diff
@@ -527,6 +529,120 @@ program
...
+        updates.push({
+          level: `account [${accountNumber}] in ${periodKey}`,
+          properties,
+          apply: async () => {
+            const account = await SF.findAccountByNumber(firmId, companyId, targetPeriodId, accountNumber);
+            if (!account) {
+              consola.error(`Account "${accountNumber}" not found for period ${periodKey}`);
+              return null;
+            }
+            return SF.updateAccountCustom(firmId, companyId, targetPeriodId, account.account.id, properties);
```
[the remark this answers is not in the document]
Benjvandam: Fixed

lib/utils/textPropertyUtils.js
```diff
@@ -0,0 +1,128 @@
...
+    if (keyParts.length === 2) {
+      if (!namespaceMap.has(namespaceKey)) {
+        namespaceMap.set(namespaceKey, { namespace, key, value });
+      }
+    } else {
+      if (!namespaceMap.has(namespaceKey)) {
+        namespaceMap.set(namespaceKey, { namespace, key, value: {} });
+      }
+      const subKey = keyParts.slice(2).join(".");
+      namespaceMap.get(namespaceKey).value[subKey] = value;
```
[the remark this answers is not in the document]
Benjvandam: 4 nesting levels doesn't happen in practice

lib/utils/textPropertyUtils.js
```diff
@@ -0,0 +1,128 @@
...
+    if (!fs.existsSync(testsDir)) continue;
+
+    const yamlFiles = fs.readdirSync(testsDir).filter((f) => f.endsWith(".yml"));
+
+    for (const file of yamlFiles) {
+      const filePath = path.join(testsDir, file);
+      const content = fs.readFileSync(filePath, "utf-8");
+      const parsed = yaml.parse(content, { maxAliasCount: 10000, merge: true });
+
+      if (!parsed || !parsed[testName]) continue;
```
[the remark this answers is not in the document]
Benjvandam: That is why it is possible to add the handle to the command. The first test that matches the description will otherwise be used

bin/cli.js
```diff
@@ -527,6 +529,120 @@ program
...
+      if (!response) continue;
+      // Handle both single response (reconciliation) and array of responses (company/period/account)
+      const responses = Array.isArray(response) ? response : [response];
+      const failed = responses.filter((r) => !r || r.status < 200 || r.status >= 300);
+      if (failed.length === 0) {
+        consola.success(`${update.level}: updated`);
+      } else {
+        consola.error(`${update.level}: ${failed.length}/${responses.length} failed`);
+      }
+    }
```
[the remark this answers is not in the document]
Benjvandam: Fixed

lib/utils/textPropertyUtils.js
```diff
@@ -0,0 +1,131 @@
...
+      if (!namespaceMap.has(namespaceKey)) {
+        namespaceMap.set(namespaceKey, { namespace, key, value });
+      }
+    } else {
+      if (!namespaceMap.has(namespaceKey)) {
+        namespaceMap.set(namespaceKey, { namespace, key, value: {} });
+      }
+      const subKey = keyParts.slice(2).join(".");
+      namespaceMap.get(namespaceKey).value[subKey] = value;
+    }
```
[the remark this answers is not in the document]
Benjvandam: Shouldn't happen in the YAML files in practice

bin/cli.js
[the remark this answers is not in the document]
michieldegezelle: **Stale:** This was fixed. The current code sets `hadFailures = true` throughout the loop and calls `process.exitCode = 1` after it. No action needed.

bin/cli.js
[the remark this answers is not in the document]
michieldegezelle: **Note:** This doesn't apply to the current code. The guard is `const accountId = account?.account?.id` — if `findAccountByNumber` returns a truthy error object (e.g. from `responseErrorHandler`), the optional chain evaluates to `undefined` because the error object has no `.account` property, so `!accountId` catches it correctly. No change needed.

bin/cli.js
```diff
@@ -527,6 +529,129 @@ program
...
+      updates.push({ level: "company", properties, apply: () => SF.updateCompanyCustom(firmId, companyId, properties) });
+    }
+
+    // Fetch periods once if needed for resolving period dates
+    let periodsArray = null;
+
+    for (const [periodKey, periodEntry] of Object.entries(testData.periods)) {
+      // Resolve period date to period ID
+      if (!periodsArray) {
+        const periodsResponse = await SF.getPeriods(firmId, companyId);
```
michieldegezelle: **🟠 Major:** `getPeriods` fetches only the first page (200 periods). Companies with more than 200 periods will silently skip all updates past that page — the loop falls through to the `consola.warn('Period not found')` branch with exit code 0, giving no indication that data was missed. The existing `getAllPeriodCustom` pattern (page loop until `pageData.length < PER_PAGE`) should be applied here:

```js
let periodsArray = [];
let page = 1;
while (true) {
  const resp = await SF.getPeriods(firmId, companyId, page);
  const batch = resp?.data || [];
  periodsArray = periodsArray.concat(batch);
  if (batch.length < 200) break;
  page++;
}
```

bin/cli.js
```diff
@@ -527,6 +529,129 @@ program
...
+        periodsArray = periodsResponse?.data || [];
+      }
+      const period = periodsArray.find((p) => p.fiscal_year.end_date === periodKey);
```
michieldegezelle: **🟡 Minor:** `p.fiscal_year` can be `null` for certain period types in the Silverfin API. Accessing `.end_date` directly throws a TypeError; use optional chaining:

```js
const period = periodsArray.find((p) => p.fiscal_year?.end_date === periodKey);
```

lib/utils/textPropertyUtils.js
```diff
@@ -0,0 +1,128 @@
...
+  for (const dir of handleDirs) {
+    const testsDir = path.join(baseDir, dir, "tests");
+    if (!fs.existsSync(testsDir)) continue;
+
+    const yamlFiles = fs.readdirSync(testsDir).filter((f) => f.endsWith(".yml"));
+
+    for (const file of yamlFiles) {
+      const filePath = path.join(testsDir, file);
+      const content = fs.readFileSync(filePath, "utf-8");
+      const parsed = yaml.parse(content, { maxAliasCount: 10000, merge: true });
```
michieldegezelle: **🟡 Minor:** `yaml.parse` is not wrapped in a try/catch. A malformed YAML file in any reconciliation's `tests/` directory crashes the command with a raw stack trace instead of a user-friendly skip/error. The existing `liquidTestUtils` parsing wraps `yaml.parse` similarly — worth doing the same here:

```js
let parsed;
try {
  parsed = yaml.parse(content, { maxAliasCount: 10000, merge: true });
} catch (e) {
  consola.warn(`Skipping ${filePath}: invalid YAML — ${e.message}`);
  continue;
}
if (!parsed || !parsed[testName]) continue;
```

michieldegezelle (commented):
**🟠 Major — Missing test file: `lib/utils/textPropertyUtils.js`**

Every other file in `tests/lib/utils/` has a corresponding `*.test.js` (`liquidTestUtils.test.js`, `urlHandler.test.js`, etc.), but `textPropertyUtils.js` has none. The two functions have non-trivial logic worth covering:

- `transformCustomToProperties`: key splitting, silent-drop behaviour for duplicate/conflicting dotted keys, mixed-depth keys
- `findTestData`: multi-directory scan, first-match selection, edge cases on missing test name

Without tests, regressions in parsing logic will only surface when users run the real command against a live company.

lib/utils/textPropertyUtils.js
```diff
@@ -0,0 +1,128 @@
...
+const yaml = require("yaml");
+const { consola } = require("consola");
+const fsUtils = require("./fsUtils");
+
+/**
+ * Transform a YAML custom properties object into the Silverfin API format.
+ * Input: flat object with dot-notation keys (e.g. "namespace.key.subkey": value)
+ * Output: array of { namespace, key, value } objects
+ */
+function transformCustomToProperties(customData) {
```
michieldegezelle, line 12: **💡 Suggestion:** This module contains non-trivial transformation and YAML lookup logic (`transformCustomToProperties`, `findTestData`) but has no unit tests. Consider adding `tests/lib/textPropertyUtils.test.js` covering dot-notation transforms, nested-key edge cases, and `findTestData` with `--handle` scoping — similar to other `lib/utils` modules in this repo.

bin/cli.js
```diff
@@ -527,6 +529,129 @@ program
     liquidTestGenerator.testGenerator(options.url, testName, reconciledStatus);
   });
 
+// Update Text Properties from Liquid Test data
+program
+  .command("update-text-properties")
+  .description("Upload custom text properties from a Liquid Test YAML file to a reconciliation in a company file")
```
michieldegezelle: **🟠 Major:** The command description says properties are uploaded "to a reconciliation in a company file", but the implementation pushes custom data at company, period, reconciliation, and account levels for every entry in the test YAML. That mismatch will surprise users who paste a reconciliation URL expecting a single-target update.

Either narrow the scope to the URL's period/reconciliation (as CodeRabbit suggested) or update the description/CHANGELOG to state explicitly that this seeds the full test scenario across all YAML levels.

bin/cli.js
Benjvandam: Fixed in `getAllPeriods()` (lib/api/sfApi.js), which paginates the periods endpoint the same way as `getAllPeriodCustom` (loops pages until a short page, capped at MAX_PAGES with a warning). The command now resolves periods across all pages instead of only the first 200.

bin/cli.js
Benjvandam: Fixed — the period match now uses optional chaining: `p.fiscal_year?.end_date === periodKey`.

lib/utils/textPropertyUtils.js
Benjvandam: Fixed — `yaml.parse` is now wrapped in try/catch; a malformed YAML file is skipped with a `consola.warn` rather than crashing the command.

lib/utils/textPropertyUtils.js
Benjvandam: Added `tests/lib/utils/textPropertyUtils.test.js` (9 cases) covering `transformCustomToProperties` (simple/nested/merge/skip) and `findTestData` (handle scoping, folder scan, all-level extraction, malformed-file skip, and not-found).

bin/cli.js
Benjvandam: Kept the intentional multi-level behaviour — the command sets up a full test scenario, not a single record. To address the surprise factor: the description now states it writes at the company/period/reconciliation/account levels for every entry in the test, and before uploading it prints the target firm plus per-level property counts and asks for confirmation (-y/--yes to skip).

bin/cli.js
```diff
@@ -527,6 +529,145 @@ program
...
+    let periodsArray = null;
+
+    for (const [periodKey, periodEntry] of Object.entries(testData.periods)) {
+      // Resolve period date to period ID
+      if (!periodsArray) {
+        periodsArray = await SF.getAllPeriods(firmId, companyId);
+      }
+      const period = periodsArray.find((p) => p.fiscal_year?.end_date === periodKey);
+      if (!period) {
+        consola.warn(`Period "${periodKey}" not found in company — skipping`);
```
michieldegezelle: 🟠 **Major:** A period listed in the test YAML but absent from the target company is only `consola.warn`-ed and skipped, with no failure flag set — so the command still exits `0`. This is inconsistent with the reconciliation/account not-found paths below (which return `null` → set `hadFailures` → `process.exitCode = 1`). A scenario that is only partially seeded — or, if `getAllPeriods` returns `[]` because the periods endpoint errored, not seeded at all (the `updates.length === 0` branch then reports "No custom properties found" and exits `0`) — looks green to CI and scripts.

```js
      const period = periodsArray.find((p) => p.fiscal_year?.end_date === periodKey);
      if (!period) {
        consola.error(`Period "${periodKey}" not found in company — its custom data will NOT be written`);
        hadFailures = true; // declare `let hadFailures = false;` before this loop and OR it into the final process.exitCode check
        continue;
      }
```

lib/api/sfApi.js
```diff
@@ -572,6 +600,66 @@ async function getAllPeriodCustom(firmId, companyId, periodId) {
   return items;
 }
 
+async function updateReconciliationCustom(firmId, companyId, periodId, reconciliationId, properties) {
+  const instance = AxiosFactory.createInstance("firm", firmId);
+  try {
+    const response = await instance.post(`/companies/${companyId}/periods/${periodId}/reconciliations/${reconciliationId}/custom`, { properties });
```
michieldegezelle, line 612: 🟠 **Major:** `updateReconciliationCustom` POSTs the whole set as a single bulk body `{ properties: [...] }`, whereas `updateCompanyCustom` / `updatePeriodCustom` / `updateAccountCustom` (added in the same PR, against the same `/custom` sub-resource) POST one `{ namespace, key, value }` object per request in a loop. Please confirm the reconciliation `/custom` endpoint actually accepts the bulk `{ properties }` shape — if it does not, the write returns 422 and `apiUtils.responseErrorHandler` calls `process.exit(1)`, hard-aborting the whole command mid-batch. If bulk is not supported, mirror the per-property loop used by the other three; if it is supported, consider using the bulk shape for the other three as well for consistency and fewer round-trips.

bin/cli.js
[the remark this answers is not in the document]
michieldegezelle: **Note:** This is intentional per the updated command description (lines 535–537): the URL identifies firm/company, and the YAML defines the full scenario to seed across company/period/reconciliation/account levels. Not a defect — closing the loop on the original concern.

bin/cli.js
```diff
@@ -527,6 +529,145 @@ program
...
+  .option("-h, --handle <handle>", "Specify the reconciliation handle to narrow down the YAML file search (optional)")
+  .option("--dry-run", "Only transform and display the properties without uploading (optional)", false)
+  .option("-y, --yes", "Skip the confirmation prompt before uploading (optional)", false)
+  .action(async (options) => {
+    // Parse URL to extract IDs
+    const urlData = liquidTestUtils.extractURL(options.url);
+    const { firmId, companyId } = urlData;
+    if (!firmId || !companyId) {
+      consola.error("Could not determine the firm and company from the URL. Double-check the Silverfin URL and try again.");
+      return;
```
michieldegezelle, line 551: **🟡 Minor:** When `firmId` or `companyId` cannot be parsed from the URL, this branch logs an error and `return`s, so the process still exits with code 0. Use `process.exit(1)` (or set `process.exitCode = 1` before returning) so scripts/CI can detect invalid input — consistent with the `hadFailures` handling later in this command.

lib/api/sfApi.js
```diff
@@ -572,6 +600,66 @@ async function getAllPeriodCustom(firmId, companyId, periodId) {
...
+async function updateCompanyCustom(firmId, companyId, properties) {
+  const instance = AxiosFactory.createInstance("firm", firmId);
+  const results = [];
+  for (const prop of properties) {
+    try {
+      const response = await instance.post(`/companies/${companyId}/custom`, prop);
+      apiUtils.responseSuccessHandler(response);
+      results.push(response);
+    } catch (error) {
+      const response = await apiUtils.responseErrorHandler(error);
```
michieldegezelle: **🟠 Major:** `apiUtils.responseErrorHandler` calls `process.exit(1)` on HTTP 422. If any single property POST in `updateCompanyCustom` / `updatePeriodCustom` / `updateAccountCustom` returns 422, the entire CLI aborts mid-batch — the remaining updates in the `updates` loop never run and `hadFailures` is bypassed. Consider handling 422 like other errors (return the error response to the caller) so the command can finish the batch and exit via `process.exitCode = 1` at the end.

bin/cli.js
Benjvandam: Fixed in aeddf10: a period listed in the YAML but absent from the company now logs an error and sets `process.exitCode = 1` (the remaining scenario still seeds, but the partial seed is detectable) — consistent with the reconciliation/account not-found paths. `getAllPeriods` returning `[]` hits the same path, since every period lookup then fails.

lib/api/sfApi.js
Benjvandam: Good catch — settled it empirically against the live API (test writes with a throwaway namespace, cleaned up after): the **period**, **reconciliation** and **account** `/custom` endpoints all accept the bulk `{ properties: [...] }` body (201; the reconciliation one is also proven by live use — a bulk write recomputed results). The **company** `/custom` endpoint does **not**: it returns 400 (`namespace is missing, key is missing, value is missing`) unless each custom is POSTed as a single `{ namespace, key, value }` object. So in aeddf10 I aligned `updatePeriodCustom` and `updateAccountCustom` to the single bulk POST, and kept the per-property loop only for `updateCompanyCustom`, with a comment documenting the API difference.

bin/cli.js
Benjvandam: Fixed in aeddf10: the unparseable-URL branch now sets `process.exitCode = 1` before returning.

lib/api/sfApi.js
Benjvandam: Fixed in aeddf10: the four `update*Custom` functions now use a new `apiUtils.batchResponseErrorHandler`, which logs and returns the error response instead of `process.exit(1)` on 422/403 — so a failed write is counted via `hadFailures` and the remaining updates in the batch still run. I added a separate handler rather than changing `responseErrorHandler`, since that one is shared by many non-batch commands where aborting is the intended behavior. (Network-level errors without an HTTP response still rethrow — those aren't per-item failures.)

bin/cli.js
michieldegezelle: **Stale:** This was fixed — `getAllPeriods()` in `lib/api/sfApi.js` now paginates until a short page (capped at `MAX_PAGES` with a warning), mirroring `getAllPeriodCustom`. No action needed.

bin/cli.js
michieldegezelle: **Stale:** Resolved — the command description (line 536) now accurately states it writes at company/period/reconciliation/account levels for every entry in the test scenario, matching the implementation. No action needed.

bin/cli.js
michieldegezelle: **Stale:** This was fixed — a period listed in the YAML but absent from the company now logs an error and sets `process.exitCode = 1` (bin/cli.js:579-580), consistent with the reconciliation/account not-found paths. No action needed.

lib/api/sfApi.js
michieldegezelle: **Note:** Accepted based on the author's live verification (bulk `{ properties: [...] }` accepted on period/reconciliation/account `/custom`, rejected on company `/custom`). The now-documented comment above `updateReconciliationCustom` (lib/api/sfApi.js:603-608) captures the rationale for future readers. No action needed.

lib/api/sfApi.js
michieldegezelle: **Stale:** This was fixed — the four `update*Custom` functions now use `apiUtils.batchResponseErrorHandler`, which logs and returns the error response on 422/403 instead of calling `process.exit(1)`, so `hadFailures` correctly tracks the failure and the remaining batch continues. No action needed.

bin/cli.js
```diff
@@ -527,6 +529,150 @@ program
...
+    }
+    if (!options.yes) {
+      cliUtils.promptConfirmation();
+    }
+
+    // Apply all updates
+    let hadFailures = false;
+    for (const update of updates) {
+      consola.start(`Updating ${update.level} (${update.properties.length} properties)...`);
+      const response = await update.apply();
```
michieldegezelle: 🟠 **Major:** `await update.apply()` isn't wrapped in a try/catch, so a thrown error from the reconciliation/account **lookup** helpers aborts the entire remaining batch — the same failure mode this PR just fixed for the write calls via `batchResponseErrorHandler`.

Concretely, `apply()` for reconciliation/account updates (lines 597-598, 614-615) calls `SF.findReconciliationInWorkflows` / `SF.findAccountByNumber`, which still resolve through the old `apiUtils.responseErrorHandler`:
- On 403/422 it calls `process.exit(1)` directly — killing the whole command mid-batch.
- On any other unhandled status (401, 5xx) it rethrows, which isn't caught anywhere in this loop, so it becomes an uncaught exception (`errorUtils.uncaughtErrors` → `process.exit(1)`).
- If `getWorkflows` (called from `findReconciliationInWorkflows`) returns an error response instead of throwing (e.g. 404), `responseWorkflows.data` is `undefined`, and `for (const workflow of responseWorkflows.data)` throws a TypeError.

Any of these mid-batch turns what should be one failed `update` (tracked via `hadFailures`) into a hard crash that silently drops every remaining update in the run — undermining the batch-continuation goal this PR just implemented for the direct write path.

Suggested fix — wrap the resolve+apply step so lookup failures are treated the same as write failures:
```js
for (const update of updates) {
  consola.start(`Updating ${update.level} (${update.properties.length} properties)...`);
  let response;
  try {
    response = await update.apply();
  } catch (error) {
    hadFailures = true;
    consola.error(`${update.level}: failed to resolve/update — ${error.message}`);
    continue;
  }
  if (!response) {
    hadFailures = true;
    continue;
  }
  ...
}
```

lib/utils/textPropertyUtils.js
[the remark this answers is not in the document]
Benjvandam: Reworked in the latest commits — the command no longer takes the first match. It now reads the test file referenced by the template's config.json `test` key (the same file `run-test` uses), and a new `--file <exact-file-name>` option selects a specific variant (e.g. `*_TY25_liquid_test.yml`). If the test name exists in multiple templates it lists the candidates and asks for `--handle` instead of guessing.

bin/cli.js
Benjvandam: Fixed — the apply loop now wraps `update.apply()` in a try/catch: a thrown error (network failure, lookup 500) is logged, counted as a failure (exit code 1) and the batch continues with the remaining updates. Also guarded `findReconciliationInWorkflows` against an undefined workflows response, so a handled 404 warns "not found" instead of throwing.

lib/utils/textPropertyUtils.js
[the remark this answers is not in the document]
michieldegezelle: **Note:** Confirmed unchanged at HEAD — the mixing-keys behavior described here still doesn't occur in practice per the author's earlier reply, and this is already recorded as an accepted convention for this file. No action needed.

lib/utils/textPropertyUtils.js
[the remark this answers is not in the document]
michieldegezelle: **Note:** Confirmed unchanged at HEAD — 4+ nesting levels still don't occur in practice per the author's earlier reply, and this is already recorded as an accepted convention for this file. No action needed.

lib/utils/textPropertyUtils.js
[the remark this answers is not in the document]
michieldegezelle: **Stale:** This was superseded. `findTestData` no longer takes the first YAML match across all templates — it now reads the test file referenced by the template's `config.json` `test` key (confirmed at HEAD), with `--file` for selecting a specific variant. No action needed.

bin/cli.js
michieldegezelle: **Stale:** This was fixed. `update.apply()` is now wrapped in try/catch in the apply loop (confirmed at HEAD) — a thrown error is logged and counted via `hadFailures`. No action needed.

bin/cli.js
```diff
@@ -527,6 +529,160 @@ program
...
+      updates.push({ level: "company", properties, apply: () => SF.updateCompanyCustom(firmId, companyId, properties) });
+    }
+
+    // Fetch all periods once (paginated) if needed for resolving period dates
+    let periodsArray = null;
+
+    for (const [periodKey, periodEntry] of Object.entries(testData.periods)) {
+      // Resolve period date to period ID
+      if (!periodsArray) {
+        periodsArray = await SF.getAllPeriods(firmId, companyId);
```
michieldegezelle, line 573: 🟠 **Major:** `SF.getAllPeriods(firmId, companyId)` isn't wrapped in try/catch here — a network error escapes every per-update failure handler in this loop and aborts the whole command, discarding the `hadFailures` accounting already collected for prior updates (company-level, and any earlier periods). Convergent finding from 2 independent finders.

```js
if (!periodsArray) {
  try {
    periodsArray = await SF.getAllPeriods(firmId, companyId);
  } catch (error) {
    consola.error(`Could not fetch periods: ${error.message}`);
    process.exitCode = 1;
    return;
  }
}
```

lib/api/sfApi.js
```diff
@@ -572,6 +600,60 @@ async function getAllPeriodCustom(firmId, companyId, periodId) {
...
+async function updateCompanyCustom(firmId, companyId, properties) {
+  const instance = AxiosFactory.createInstance("firm", firmId);
+  const results = [];
+  for (const prop of properties) {
+    try {
+      const response = await instance.post(`/companies/${companyId}/custom`, prop);
+      apiUtils.responseSuccessHandler(response);
+      results.push(response);
+    } catch (error) {
+      results.push(apiUtils.batchResponseErrorHandler(error));
```
michieldegezelle, line 629: 🟡 **Minor:** `batchResponseErrorHandler` rethrows on a non-HTTP error (network/timeout), which escapes this `for` loop — any properties already collected in `results` for this company update are discarded. It's still caught one level up in `bin/cli.js`'s per-update try/catch, so the command doesn't crash, but this one company-level update loses its per-property success/failure granularity in the report. Convergent finding from 2 independent finders.

```js
for (const prop of properties) {
  try {
    const response = await instance.post(`/companies/${companyId}/custom`, prop);
    apiUtils.responseSuccessHandler(response);
    results.push(response);
  } catch (error) {
    try {
      results.push(apiUtils.batchResponseErrorHandler(error));
    } catch (rethrown) {
      results.push({ status: 0, statusText: rethrown.message });
    }
  }
}
return results;
```

delphine-demeulenaere (changes requested):
**General feedback:**

Reviewed the code and did functional testing on my end. Two things still need attention before this is ready to merge:

- There are 2 open code comments with pending implementation. Both fixes are needed to ensure failures mid-run are handled gracefully, keeping the process going and reporting errors cleanly, rather than crashing or losing progress.
- Merge conflicts need to be resolved.
- Functional testing, tested against a BE market template (PLT templates don't have liquid tests)

Simple liquid test: works correctly. Template is updated with the liquid test text properties as expected.
Image

Extensive liquid test (with company, period, recon and account customs): also working correctly. When a period doesn't exist in the client file, it's flagged as expected. All other updates are applied correctly to the liquid test text properties in the template. 
Image

lib/api/sfApi.js
delphine-demeulenaere: Implemented above code

bin/cli.js
delphine-demeulenaere: Implemented above code

PR #260, opened 2026-06-29
Live bridge: capture, understand, manipulate & verify a company file over the v4 API

Benjvandam:
## What this is

The **live bridge**: drive the full **capture → understand → manipulate → verify → push** loop on a live Silverfin company file over the v4 API — no browser. Change inputs and *see the recomputed results* over HTTP, cheaply, and iterate.

This PR grew from the original 5 commands into the full feature and is the **hub**. Two companion PRs land with it:
- **silverfin-ls** — silverfin/silverfin-ls#1 — the `data-scope` command this CLI delegates static analysis to.
- **bso-cursor-rules** — silverfin/bso-cursor-rules#37 — the `silverfin-live-bridge` skill that teaches the agent the loop.

## How it fits together

**1. Understand (static, via silverfin-ls)**
- `manifest` — a template's data scope (own customs, cross-template deps incl. dynamically-keyed, prior-period depth, company drop, accounts, involved files). Delegates to `silverfin-ls data-scope`, reusing its maintained tree-sitter parser instead of duplicating STL parsing.
- `describe-inputs` — every input with type, declared default, stored value and live **effective** value, labelling the source (stored override vs. default vs. result-echo). It resolves inputs that are echoed as a result — **directly** *or* **indirectly** (a result echoing a variable derived from the input) — straight from the live results, **with no silverfin-ls needed**. Only the residual (defaults not surfaced as a result) needs the steps below.
- **Inspect a sibling/upstream template** (the ones that feed another's defaults) with `--handle <handle>` on `get-results` / `describe-inputs` / `capture` — same company/period, no URL to reconstruct — or `resolve-handle` for its instance id + URL. `describe-inputs --handle` also surfaces **line-item collections** (`{% fori item in custom.items %}…`) as a `collections` section with the live dated line-items, so collection-driven inputs (e.g. a working paper's amounts) are visible.

**2. Capture (live — the oracle)**
- `get-results` — computed results + customs (the API returns them directly).
- `capture` / `capture --full` — JSON snapshot of a template's scope, or the whole company across all periods/reconciliations, from firm+company alone.

**3. Resolve the residual hidden defaults (optional specialist)**
For defaults that *aren't* surfaced as a result: `describe-inputs --resolve` fills direct cross-template references from a targeted capture; `--compute` runs a **bounded, never-fabricating offline STL evaluator** (`lib/stlLite.js`) for computed / dynamic-key / account-range defaults, labelled `computed:… (validate vs live)`. The platform stays the source of truth; this is the browser-free specialist.

**4. Manipulate — three deliberate modes**
- `set-custom` / `delete-custom` — override a field (any level), `--dry-run`, `--handle` to target an underlying template, `--file` batch.
- **`set-default`** — change a field's *default* by writing the **upstream custom it derives from** (reverse-traced via `lib/provenanceTracer.js`), never an override. Auto-proceeds when auto-invertible, prints a change table (old → new + why + blast radius), and otherwise surfaces the provenance chain — never guesses.
- `update-text-properties` — bulk-seed from a Liquid Test YAML.

**5. Verify** — `get-results` again; compare (its output now includes a `resultCount`, with a `note` when 0 so an empty results table is self-explanatory — the template isn't computing in that company, *not* a caching issue). The platform is the source of truth.

**Live data vs. template logic** — for verifying what a template's *code* produces (not live-data effects), a new **`render`** command renders it against its local liquid-test fixture and returns a JSON outcome (rendered? / expectation mismatches) — deterministic, and it works even where a live company has no data (where `get-results` is empty).

## Key modules
`lib/dataScope.js` (silverfin-ls shell-out + auto-detect + `doctor`), `lib/deepCapture.js` (targeted deep capture + `period.accounts`), `lib/offlineDefaultResolver.js` + `lib/stlLite.js` (offline evaluator), `lib/defaultResolver.js` (direct-ref), `lib/provenanceTracer.js` + `lib/defaultSetter.js` + `lib/changeReport.js` (set-default), `lib/inputDescriber.js` (incl. indirect/adjacent echo resolution), `lib/customWriter.js`.

## Robustness / diagnostics
The CLI **auto-detects** silverfin-ls (`SILVERFIN_LS_CMD` → `silverfin-ls` on PATH → `npx --no-install silverfin-ls`) and puts a **timeout** on every call, so an outdated silverfin-ls (which would ignore `data-scope` and start its LSP server) can no longer hang the CLI. A new **`silverfin doctor`** command probes silverfin-ls end-to-end and reports OK or the exact failure with install steps. And because `describe-inputs` resolves echoed inputs from the live results without silverfin-ls, the tool degrades gracefully when it's absent.

## Safety
Never fabricates a value (the offline evaluator gives up rather than guess); `--dry-run` on every write; test-firm default in the skill; `set-default` was hardened via an adversarial review against value-changing filters, field-access customs and period-collisions to avoid wrong writes.

## Dependency / release order
The CLI shells out to `silverfin-ls data-scope` (those commands error clearly — run `silverfin doctor` — if it's absent). **Usable order: silverfin-ls merged + released → this CLI merged + released → the skill live.**

## Base / stacking
This branch is based on `implement-reverse-test` (#249). Once #249 merges we can retarget this PR to `main` so it shows only the live-bridge feature.

## Tests
633 jest passing, eslint clean. CHANGELOG under `[Unreleased]` (no self version bump). The silverfin-ls shell-out isn't exercised in CI (unit tests are pure), so CI is green without silverfin-ls installed.


BenjaminLangenakenSF:
Review note — scope, not merit.

**The data copier does not supersede this.** `copyCompanyData` ([index.js#L1360-L1398](https://github.com/silverfin/silverfin-cli/blob/b1991fcf931e8e2fb3b06a4e26744507d0a5f756/index.js#L1360-L1398)) is a fire-and-forget trigger for a server-side Ruby job — no status polling, produces a *new* company, never reports what a template computed. Different job. Its fidelity gaps (zero accounts, id-keyed customs, locked-period remapping) are a reason to *want* `capture --full` for source-vs-copy verification.

**The concern is size.** 6,713-line diff, 14 new `lib/` modules, a hard dependency on `silverfin-ls` (separate repo), and `stlLite.js` re-implements STL evaluation the platform owns — a permanent drift risk.

**Proposal:** split. `update-text-properties` + `get-results` + `describe-inputs` alone satisfy #242 ("Allow to push test data to company") and carry most of the debugging value. The static-analysis and offline-evaluator half (`manifest`, `set-default`, `stlLite`, `silverfin-ls`) lands separately, judged on its own.

@Benjvandam has left, so this PR needs a new owner before it can move.

PR #261, opened 2026-07-01
Surface sampler report URL and quiet CI output

michieldegezelle:
Always log the hosted result_url on completion so it can be captured in CI (e.g. echoed into $GITHUB_STEP_SUMMARY) without downloading anything. Only download+open the report locally when not in CI; add --no-open. Use a static poll message instead of the animated spinner in CI to avoid flooding the captured log with thousands of frames.

## Description

## Testing Instructions

Steps:

1. ...
2. ...
3. ...

## Author Checklist

## Reviewer Checklist


lib/liquidSamplerRunner.js
```diff
@@ -241,11 +271,17 @@ class LiquidSamplerRunner {
...
+          // (e.g. echoed into $GITHUB_STEP_SUMMARY) without downloading anything.
+          consola.success(`Sampler report: ${response.result_url}`);
+          if (this.openReport) {
+            await new UrlHandler(response.result_url).openFile();
+          }
+          if (this.compact) {
+            await this.#printCompactDiff(response.result_url);
+          }
+        } else {
+          consola.warn("No URL returned");
```
michieldegezelle: 🟠 **Major:** `completed` with no `result_url` now only `consola.warn("No URL returned")` — no `process.exit(1)` / `process.exitCode` is set anywhere in the call chain, so the CLI now exits `0` for a completed-but-urlless sampler run.

This is a real regression, not hypothetical: commit `cba14f8` ("Minor updates to error handling of handleSamplerResponse", 2026-07-14) deliberately changed this exact branch from a bare warn to `consola.error(...); process.exit(1);`. That commit is an ancestor of this PR's head (`git merge-base --is-ancestor cba14f8 HEAD` confirms it), but this PR's commit `58e7747` rewrote the same block back to warn-only — almost certainly an accidental clobber during a rebase whose patch context still matched cleanly, not an intentional design choice.

Given `run-sampler` is meant to be greppable in CI for the `Sampler report: <url>` line (per this branch's own stated goal), a completed run with no URL now produces neither the report line nor a non-zero exit — it will look identical to a quiet success.

michieldegezelle (commented):
🟠 **Major** — `lib/liquidSamplerCompact.js:145` (can't be posted as an inline comment because GitHub treats this new file as binary and has no diff positions for it — see below):

```js
const dedupKey = `${change.key}\\x00${change.before}\\x00${change.after}`;
```

This line contains a **literal raw NUL byte** (`0x00`) between the template-literal segments, not the two-character escape sequence `\\x00`. Confirmed with `file lib/liquidSamplerCompact.js` → `data` (not "ASCII text"), and `gh pr diff 261` renders this entire new 213-line file as `Binary files /dev/null and b/lib/liquidSamplerCompact.js differ` — it cannot be reviewed line-by-line on GitHub at all, and any future diff touching this file will likely render as binary too.

Functionally the code still works (a literal NUL is a valid, if unusual, dedup separator in a JS string), but this should be the actual two-character escape sequence in source so the file stays reviewable text going forward.

michieldegezelle (commented):
🟠 **Major** — `lib/liquidSamplerCompact.js:61-66` (can't be posted as an inline comment — this file has a literal NUL byte at line 145 that makes GitHub treat the whole file as binary, so it has no diff positions; see the other top-level comment on this PR about that. This finding was originally raised by CodeRabbit's first review round on this file, but that round's inline comments all failed to post due to a GitHub API error, and it was never re-surfaced in the later round either — still present and unfixed):

```js
for (const kind of ENTRY_KINDS) {
  for (const [entryId, meta] of Object.entries(parsed[kind] || {})) {
    labelMap[entryId] = {
      label: (meta && meta.label) || entryId,
      url: (meta && meta.url) || null,
    };
  }
}
```

`labelMap` is keyed only by the raw `entryId`, shared across both `account_entries` and `reconciliation_entries`. Their id spaces aren't guaranteed disjoint, so if an account entry and a reconciliation entry happen to share a numeric id, the second kind processed silently overwrites the first's label/url — and `extractCompact`'s lookup (`labelMap[entryId]`, ~line 138) can then attribute a changed entry to the wrong template. Key by `${kind}/${entryId}` in both this map and the corresponding lookup/grouping in `extractCompact`.

michieldegezelle (commented):
🟠 **Major** — `lib/liquidSamplerCompact.js:126-133` (same situation as the previous comment: raised by CodeRabbit's first review round, never actually posted due to that round's inline-comment failure, never re-surfaced afterward, still present and unfixed):

```js
for (const entryId of fs.readdirSync(kindDir)) {
  const entryDir = path.join(kindDir, entryId);
  if (!fs.statSync(entryDir).isDirectory()) continue;
  entriesSampled += 1;

  const before = readNamedResults(path.join(entryDir, "before", "registers.json"));
  const after = readNamedResults(path.join(entryDir, "after", "registers.json"));
  if (before === null || after === null) continue;
```

`entriesSampled` increments before checking whether `before`/`after` actually parsed. An entry whose `registers.json` is missing or malformed still counts as "sampled" even though it's silently skipped from the diff — so `formatCompact`'s "No `named_results` changes across N sampled entries" message can overstate how many entries were genuinely analyzed, masking entries that were never compared at all. Either count only entries where both files parsed successfully, or track and surface the skipped count separately.

lib/liquidSamplerRunner.js
michieldegezelle: **Stale:** This was fixed. No action needed.

lib/liquidSamplerCompact.js
```diff
@@ -0,0 +1,253 @@
...
+/**
+ * Read the named_results object from a registers.json file.
+ * @param {string} filePath
+ * @returns {Object|null} the named_results object, or null if unreadable
+ */
+function readNamedResults(filePath) {
+  try {
+    const registers = JSON.parse(fs.readFileSync(filePath, "utf8"));
+    const named = registers.named_results;
+    return named && typeof named === "object" ? named : {};
```
michieldegezelle, line 88: 🟡 **Minor:** `readNamedResults` only returns `null` (counted as skipped) when `JSON.parse` throws or the parsed value is null/undefined. If `registers.json` is valid JSON but not an object (e.g. truncated mid-write into a bare string/number, or an array), `registers.named_results` is `undefined`, and the function falls through to `{}` instead of `null`. That entry then counts as sampled (not skipped) with an empty before/after, so any real keys on the *other* side of the pair get reported as spurious ABSENT→value / value→ABSENT changes instead of the entry being flagged as unreadable.

lib/liquidSamplerRunner.js
```diff
@@ -258,6 +300,67 @@ class LiquidSamplerRunner {
...
+   * report URL was logged; the compact diff is a best-effort convenience.
+   * @param {string} resultUrl - Presigned URL to the results.zip
+   */
+  async #printCompactDiff(resultUrl) {
+    let resultsDir;
+    try {
+      resultsDir = await this.#downloadAndExtractResults(resultUrl);
+      const data = extractCompact(resultsDir);
+      // Plain stdout (not consola) so the markdown is captured verbatim in CI.
+      console.log(`\n${COMPACT_START}\n${formatCompact(data)}\n${COMPACT_END}`);
```
michieldegezelle: 🟡 **Minor:** `named_results` values are template-controlled and get embedded verbatim (via `renderValue`/`JSON.stringify`) between the fixed `SAMPLER_COMPACT_START`/`SAMPLER_COMPACT_END` markers. If a value's rendered form ever contains the literal marker text (e.g. a template legitimately outputs an HTML comment string), a downstream consumer doing a naive substring/regex extraction between the markers (this is explicitly the stated purpose of the markers, e.g. a CI job posting this section as a PR comment) will find the embedded fake marker first and truncate/misparse the real section.

lib/liquidSamplerRunner.js
```diff
@@ -190,7 +217,15 @@ class LiquidSamplerRunner {
     const pollingDelay = 15000; // 15 seconds
     const waitingLimit = 3600000; // 1 hour
 
-    spinner.spin("Running sampler...");
+    // The animated spinner writes a frame on every tick; without a TTY (e.g. CI)
+    // that floods the captured log with thousands of lines. Use a single static
+    // line there instead.
+    const useSpinner = !process.env.CI;
```
michieldegezelle: 🟡 **Minor:** `useSpinner = !process.env.CI` diverges from this codebase's established non-interactive detection, `lib/cli/spinner.js`'s own `!stdout.isTTY` check (introduced for the same reason in the 1.56.3 changelog entry). Two consequences: (1) in a non-CI, non-TTY invocation (piped output, cron, `nohup`), `useSpinner` is `true`, so `spinner.spin()` is called but silently no-ops on its own `!stdout.isTTY` guard — the `consola.info(...)` fallback never fires either, leaving zero polling feedback for up to the full 1-hour `waitingLimit`. (2) even in real CI, only one line is logged before the loop goes silent for up to an hour, which some CI runners' inactivity timeouts (not GitHub Actions, but others) could interpret as a hung job. Consider keying off `stdout.isTTY` (matching `spinner.js`) and/or emitting a periodic heartbeat line during the poll loop.

michieldegezelle (commented):
🟠 **Major:** `check-cli-version` is currently failing on this PR (verified via `gh pr checks 261` + the job log). `origin/main`'s `package.json` is now at `1.57.0` (merged separately via #236's own version bump), and this branch's `package.json`/`package-lock.json` (unchanged by this PR's diff) are also still at `1.57.0` — so the gate's "current branch version must be greater than main" check trips. `CHANGELOG.md`'s top entry (`## [1.57.0]`) only describes the original `run-sampler` command from #236, with no entry for what this PR actually adds (report-URL surfacing, `--no-open`, `--compact`, plus the zip-slip/timeout/collision/exit-code fixes). Needs a version bump (e.g. `1.57.1`) in both `package.json` and `package-lock.json`, plus a new `## [1.57.1]` `CHANGELOG.md` entry — or check "Skip bumping the CLI version" in the PR description if that's intentional.

lib/liquidSamplerCompact.js
michieldegezelle: **Stale:** Fixed in `b9f0a9e` — `readNamedResults` now rejects non-object/array parses (`if (!registers || typeof registers !== "object" || Array.isArray(registers)) return null;`) before falling through to `named_results`. Covered by the new "treats a registers.json that parses to a non-object as unreadable" test.

lib/liquidSamplerRunner.js
michieldegezelle: **Stale:** Fixed in `91f754f` — the new `escapeMarkers()` neutralizes any literal occurrence of `COMPACT_START`/`COMPACT_END` text embedded in template-controlled values before wrapping the real markers around the body, so a naive marker-delimited extractor can no longer be fooled by injected marker-like content. Covered by the new "neutralizes literal marker text embedded in a named_results value" test (asserts exactly one real start/end marker survives).

lib/liquidSamplerRunner.js
michieldegezelle: **Stale:** Fixed in `bea5ab8` — `useSpinner` now keys off `process.stdout.isTTY` (matching `lib/cli/spinner.js`'s own guard) instead of `!process.env.CI`, and a periodic heartbeat line (`heartbeatInterval`, every 60s) was added for the non-interactive path so a long poll no longer goes fully silent. Covered by the new TTY-based spinner tests and the heartbeat test.

Toby-Masters-SF (approved):
LGTM

PR #264, opened 2026-07-09
Add company-data-copier command

### LINKED ISSUE #250: Introduce data copier
*Please keep in mind that this repository, together with the submitted issues, is public.*

**Additional context**

BenjaminLangenakenSF:
## What

Adds a new `silverfin company-data-copier` CLI command that triggers the platform's **Data Copier** to copy a source company's data into a brand-new company in a destination (development) firm. This lets BSO developers reproduce a client's situation in a dev firm to debug templates against realistic data, without ever touching the production firm.

```bash
silverfin company-data-copier -c <source_company_id> -l <period_id_1> <period_id_2> -f <destination_firm_id>
```

- `-c, --source-company-id` — company to copy data from (in the source/client firm)
- `-l, --source-ledger-ids` — one or more period ids, space-separated (variadic, matching `run-test`'s `--handle`)
- `-f, --firm` — destination firm whose **token** authenticates the request and where the copy is created (defaults to the configured default firm)

It copies **data only** (account values incl. adjustments, text properties, people/company drop, configuration) — not template code. The destination firm is taken from the URL/token; the source is identified purely by company + period ids in the body (no source-firm parameter). The platform enforces source-firm access rights on the authenticating user (see the access note below).

## How it wires up

Follows the existing layering exactly:

- `lib/api/sfApi.js` → `runCompanyDataCopier`, reusing `AxiosFactory.createInstance` (so staging Basic-auth, `access_token`, and token refresh all work unchanged).
- `index.js` → `copyCompanyData` (input validation + reporting).
- `bin/cli.js` → the `company-data-copier` command.

## ✅ Route verified on live — it is the firm-scoped v4 route

An earlier revision of this command POSTed to the public, non-firm-scoped `/api/public/v3/company_data_copier/run`, on the basis that `doc/bso_company_data_copier.md` said the firm-scoped route had been renamed away (commit `24e2547bc9`). **Live probing shows that is not the case** — the copier is served at the firm-scoped route:

| Request on `live.getsilverfin.com` | Result |
|---|---|
| `POST /api/public/v3/company_data_copier/run` | **`404 page not found`** — route does not exist |
| `POST /api/v4/f/:id/company_data_copier/run` | **`401 invalid_token`** — route exists, only auth missing; returns a proper `422` business-validation once authenticated |

So this PR now POSTs the **relative** `company_data_copier/run`, inheriting the firm axios instance's `/api/v4/f/:id` baseURL (and its token, staging Basic-auth and refresh interceptor) — the "revert this one line" the previous note anticipated.

> **@team-reconciliation** — the doc's claim that `/api/v4/f/:id/company_data_copier/run` was renamed away in favour of a public-v3 route does not match the deployed `live` behaviour. Please update `doc/bso_company_data_copier.md`.

Two other alignments with the merged backend (unchanged from the previous revision):

- **Copied-company name.** The backend names the company `Company_<source_id>_<hex>` (e.g. `Company_42_a3f1b9d4`), **not** `SF_COPY_<id>_<timestamp>`. Success message + README reflect this.
- **Fire-and-forget, honestly.** `202 {"status":"enqueued"}` means *enqueued*, not copied — the copy runs in a Sidekiq job and there is **no status endpoint**. Messaging says the copy was *requested*, that it runs async, to verify in the destination firm after a few minutes, and that a company name ending in `_FAILED` means the (non-retried) copy failed.

## ⚠️ Testing with real *client* data requires a support user

The feature's purpose is reproducing **client** data in a dev firm, but doing so runs into an access boundary worth flagging explicitly:

The command authenticates as the **destination-firm** user, and the backend resolves the source company *in that user's context*. Reading a company that lives in a **different (client) firm** therefore requires the authenticating user to be either:

- a **Silverfin support user** (`silverfin_support?`), which can read any firm's companies, or
- a **member of both** the source and destination firms.

A normal BSO-developer account cannot reach client firms — the OAuth authorize page only lists the internal dev/demo/testing firms the user belongs to — so attempting a client source yields `422 {"error":"source_company_id <id> does not exist"}` (the backend reports cross-firm invisibility as "does not exist"). This is almost certainly intentional given the data-protection implications of copying production client data into a dev firm.

**Open question for the team:** how are BSO developers meant to authenticate for real client-data copies? Options: a scoped support credential for this workflow, support-initiated copies, or a consent/approval flow. Note that copies **within** firms the developer already belongs to (e.g. functional-testing → demo) work today and are sufficient to validate the command itself end-to-end (`404 → 401 → 422 validation → 202 enqueued`).

## Backend prerequisites / gotchas (now in the README)

These cause `422`s or silent skips that would otherwise confuse users:

- Destination firm **must be a demo firm** (trial/customer/junk/churn → `422`).
- Source firm must **differ** from destination firm (`422`).
- All `source_ledger_ids` must belong to `source_company_id` (`422`).
- Templates are matched to the destination firm by `handle` (reconciliations) / Dutch name `name_nl` (account detail templates). If the template doesn't already exist in the destination firm, its data is **silently skipped** — so the template you're debugging must already exist there.
- Caller must be a Silverfin support user or have access to the source firm (see access note above).
- **One copy per destination firm at a time** (Redis mutex, 5-min TTL). A second run fired at the same destination while one is in flight **silently does nothing** — don't fire the command repeatedly.

## Also included

- **Removed a misleading suffix from the shared 422 error handler** (`apiUtils.js`) — a 422 isn't specifically a permissions issue, so `"You don't have the rights to update the previous parameters"` was confusing (it showed on a Data Copier "source_company_id does not exist" error). Now only the real API error is surfaced. Applies to all commands.
- README section + CHANGELOG entry, version bump to `1.57.0`.

## Testing

- sfApi HTTP layer (`axios-mock-adapter`) asserts the request hits the firm-scoped `company_data_copier/run` route (relative to the firm `/api/v4/f/:id` baseURL).
- E2E command test (`tests/bin/cli/`) asserts the success message reports *requested/enqueued* and the follow-up guidance flags the async nature + `_FAILED` signal.
- Commander wiring in `tests/bin/cli.test.js`, plus TESTS.md catalogue entries.
- Full suite green, lint clean.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -771,6 +771,33 @@ program
     await generator.generateAndOpenFile();
   });
 
+// COPY company data (Data Copier)
+program
+  .command("company-data-copier")
+  .description("Copy a source company's data into a new company in a destination (development) firm. Copies data only (account values, text properties, people/company drop, configuration) — not template code.")
+  .requiredOption("-c, --source-company-id <company-id>", "Company id to copy data from (source firm)")
+  .requiredOption("-l, --source-ledger-ids <ledger-ids...>", "One or more period ids to copy, space-separated (the id in the source company URL between 'ledgers/' and '/workflows')")
+  .requiredOption("-f, --firm <firm-id>", "Destination firm where the copied company will be created", firmIdDefault)
+  .action(async (options) => {
+    cliUtils.checkDefaultFirm(options.firm, firmIdDefault);
+
+    const sourceCompanyId = Number(options.sourceCompanyId);
+    if (!Number.isInteger(sourceCompanyId) || sourceCompanyId <= 0) {
+      consola.error(`Invalid source company id: "${options.sourceCompanyId}". It must be a positive integer.`);
+      process.exit(1);
+    }
+
+    // Commander collects the variadic option into an array of strings (e.g. "-l 123 456").
+    const sourceLedgerIds = options.sourceLedgerIds.map((value) => Number(value));
+
+    if (sourceLedgerIds.some((id) => !Number.isInteger(id) || id <= 0)) {
+      consola.error(`Invalid source ledger (period) id in: "${options.sourceLedgerIds.join(" ")}". Each id must be a positive integer.`);
+      process.exit(1);
+    }
+
+    await toolkit.copyCompanyData(options.firm, sourceCompanyId, sourceLedgerIds);
+  });
+
 // Update the CLI
 if (pkg.repository && pkg.repository.url) {
   program
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -795,7 +795,13 @@ program
       process.exit(1);
     }
 
-    await toolkit.copyCompanyData(options.firm, sourceCompanyId, sourceLedgerIds);
+    // copyCompanyData returns false on failure (e.g. a 404/400 that responseErrorHandler swallows
+    // without exiting), so the exit code has to be set here or scripts/CI read a failed copy as success.
+    const result = await toolkit.copyCompanyData(options.firm, sourceCompanyId, sourceLedgerIds);
+
+    if (!result) {
+      process.exit(1);
+    }
   });
 
 // Update the CLI
--- a/tests/TESTS.md
+++ b/tests/TESTS.md
@@ -439,7 +439,7 @@ Source: `lib/api/sfApi.js`
 | `updateAccountTemplate` | should POST to update account template and return response | Verifies that a POST to `account_templates/:id` returns the updated account template. |
 | `findAccountTemplateByName` | should find account template by name_nl | Verifies that the matching account template is returned when its `name_nl` is found in the list. |
 | `findAccountTemplateByName` | should return null when list is empty | Verifies that `null` is returned when the list API returns an empty array. |
-| `runCompanyDataCopier` | should POST to company_data_copier/run with the attributes and return the response | Verifies that the Data Copier attributes are POSTed to `company_data_copier/run` and the response is returned. |
+| `runCompanyDataCopier` | should POST to the public v3 company_data_copier/run route with the attributes and return the response | Verifies that the Data Copier attributes are POSTed to the absolute `/api/public/v3/company_data_copier/run` route (not the firm-scoped baseURL) and the response is returned. |
 | `runCompanyDataCopier` | should delegate to the error handler on failure | Verifies that a non-2xx response is routed through `responseErrorHandler`. |
 
 ---
--- a/tests/TESTS.md
+++ b/tests/TESTS.md
@@ -31,7 +31,8 @@ tests/
 │       ├── import-shared-part.test.js
 │       ├── update-shared-part.test.js
 │       ├── create-shared-part.test.js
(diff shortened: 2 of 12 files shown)


tests/TESTS.md
```diff
@@ -422,6 +439,8 @@ Source: `lib/api/sfApi.js`
 | `updateAccountTemplate` | should POST to update account template and return response | Verifies that a POST to `account_templates/:id` returns the updated account template. |
 | `findAccountTemplateByName` | should find account template by name_nl | Verifies that the matching account template is returned when its `name_nl` is found in the list. |
 | `findAccountTemplateByName` | should return null when list is empty | Verifies that `null` is returned when the list API returns an empty array. |
+| `runCompanyDataCopier` | should POST to the public v3 company_data_copier/run route with the attributes and return the response | Verifies that the Data Copier attributes are POSTed to the absolute `/api/public/v3/company_data_copier/run` route (not the firm-scoped baseURL) and the response is returned. |
```
michieldegezelle: 💡 **Suggestion:** This catalogue entry still describes the old public-v3 absolute URL, but \`sfApi.test.js\` and the implementation now use the firm-scoped relative \`company_data_copier/run\` path — please align the test name and description with the current route.

bin/cli.js
```diff
@@ -771,6 +771,33 @@ program
...
+
+    // Commander collects the variadic option into an array of strings (e.g. "-l 123 456").
+    const sourceLedgerIds = options.sourceLedgerIds.map((value) => Number(value));
+
+    if (sourceLedgerIds.some((id) => !Number.isInteger(id) || id <= 0)) {
+      consola.error(`Invalid source ledger (period) id in: "${options.sourceLedgerIds.join(" ")}". Each id must be a positive integer.`);
+      process.exit(1);
+    }
+
+    await toolkit.copyCompanyData(options.firm, sourceCompanyId, sourceLedgerIds);
```
michieldegezelle: 🟠 **Major:** \`copyCompanyData\` returns \`false\` when the API yields no response data (404/400 path via \`responseErrorHandler\`), but this action ignores the return value so the process still exits 0 — scripts/CI will treat a failed copy request as success.

```javascript
const result = await toolkit.copyCompanyData(options.firm, sourceCompanyId, sourceLedgerIds);
if (result === false) {
  process.exit(1);
}
```

tests/TESTS.md
BenjaminLangenakenSF: Implemented with [Align TESTS.md catalogue entry with the firm-scoped Data Copier route](https://github.com/silverfin/silverfin-cli/pull/264/commits/5225462eff89cd74f4b043de1c74d84020db8213)

bin/cli.js
BenjaminLangenakenSF: Implemented with [Exit non-zero when the Data Copier request fails](https://github.com/silverfin/silverfin-cli/pull/264/commits/ada0a53fd1582a12f834e533aec9feae9fa5a104)

delphine-demeulenaere (approved):

PR #265, opened 2026-07-22
Expand run-sampler --compact: results register, vanished-output grouping, scope tier, visual-only tier

michieldegezelle:
## Summary

**DO NOT MERGE UNTIL NL AND LU PILOT PHASE HAS FINISHED**

Follow-up to #236/#261's `--compact` mode, driven by real feedback while reviewing PR #894's sampler run against nl_market: the `named_results`-only diff missed real signal and got unreadable fast (78KB/~19.7K tokens on a 202-entry NL run). This PR:

- Diffs the `results` register alongside `named_results` - 0/1 flag arrays render as a triggered-indicator count (e.g. `1/1 triggered`), other shapes (raw numeric values) fall back to a plain value diff.
- Groups entries whose `named_results`/`results` collapsed entirely (≥3 keys lost at once, all to `undefined`) into a single **"output vanished"** finding per template - previously this showed as dozens of individual "value → undefined" lines and could dominate the whole summary.
- Adds a **scope/dependency tier** (`dependencies`/`rollforward_params`/`required_keys_missing`), kept separate from the data diff so a dependency change is never mistaken for a data regression. `dependencies` renders as one sub-line per category (ledgers/handles/account ranges/company.attributes) instead of a semicolon-packed single line.
- Adds a **visual-only tier**: entries where `view.html` changed but the data diff found nothing to explain it - a rendering-only regression the old diff couldn't see at all. Described field-by-field via Silverfin's `data-name` attribute where one anchors the change, with an honest "compare the files directly" fallback where it doesn't.
- Truncates long values (accounting-policy paragraphs, notes) and caps per-template/per-entry change lists, always disclosing what was elided rather than silently dropping it.
- Every finding now links to a concrete follow-up: the entry's live app URL when known, else its path inside the results directory.
- Adds `run-sampler --from-zip <path>`: build the compact diff from an already-downloaded `results.zip`, with no sampler run or network call - lets a reviewer re-analyze a real result (or iterate on the compact-diff format itself) without a 30-60 min re-run against the shared staging backend.

Net effect measured on the real PR #894 zip (202 entries, 6 templates changed + 21 collapsed + 9 visual-only): output went from 78KB/~19.7K tokens (old format) to ~31.7KB while covering strictly more signal.

## Why keeping this open for now

This is currently pointed at from nl_market's and lu_market's `run_sampler.yml` (temporarily, on this branch) so we can gather real feedback across their approved PRs before merging - see the follow-up in each of those repos.

## Test plan
- [x] `npm test` - 640/640 passing
- [x] `npm run lint` - clean
- [x] Validated against a real downloaded `results.zip` from nl_market PR #894 via `--from-zip`, output reviewed in full
- [ ] Real end-to-end validation via a live download URL (only the local-zip path has been exercised against real data so far; the live-download path is covered by mocked tests)
- [ ] Feedback gathered from nl_market/lu_market usage before merge

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.59.0] (05/08/2026)
+Expand `silverfin run-sampler --compact`
+
 ## [1.58.0] (31/07/2026)
 Added the `company-data-copier` command, which triggers the platform Data Copier to copy a source company's data (account values incl. adjustments, text properties, people/company drop and configuration) into a brand-new company in a destination development firm. Intended for BSO developers to reproduce a client's situation in a dev firm without touching the production firm. Only *data* is copied, not template *code* — templates must already exist in the destination firm to be populated.
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 All notable changes to this project will be documented in this file.
 
 ## [1.59.0] (05/08/2026)
-Expand `silverfin run-sampler --compact`
+Expand `silverfin run-sampler --compact`: diff the `results` register, group vanished output per template instead of listing every entry, add a scope-change tier (dependencies/rollforward/required keys) and a visual-only tier for `view.html` changes the data diff can't see, and truncate/cap long output. Also adds `run-sampler --from-zip <path>` to build the compact diff from an already-downloaded `results.zip`.
 
 ## [1.58.0] (31/07/2026)
 Added the `company-data-copier` command, which triggers the platform Data Copier to copy a source company's data (account values incl. adjustments, text properties, people/company drop and configuration) into a brand-new company in a destination development firm. Intended for BSO developers to reproduce a client's situation in a dev firm without touching the production firm. Only *data* is copied, not template *code* — templates must already exist in the destination firm to be populated.
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -530,14 +530,31 @@ program
       "accountTemplate",
       "sharedPart",
       "firmIds",
+      "fromZip",
+    ])
+  )
+  .addOption(
+    new Option("--from-zip <path>", "Build the compact diff from an already-downloaded results.zip - no sampler run, no network call").conflicts([
+      "handle",
+      "accountTemplate",
+      "sharedPart",
+      "firmIds",
+      "id",
     ])
   )
   .option("--no-open", "Do not download/open the report locally; only print its URL (default in CI)")
-  .option("--compact", "Download the result and print a compact named_results diff (grouped by template) to stdout - review-friendly and safe in CI")
+  .option("--compact", "Download the result and print a compact diff (named_results/results, dependencies, vanished renders, visual-only changes) grouped by template - review-friendly and safe in CI")
   .action(async (options) => {
     // Commander sets options.open = false when --no-open is passed.
     // In CI, never open regardless of the flag.
     const runnerOptions = { openReport: options.open && !process.env.CI, compact: options.compact || false };
+
+    // A local zip needs no partner API access at all - it's pure offline
+    // re-analysis of a result someone already has on disk.
+    if (options.fromZip) {
+      await new LiquidSamplerRunner(options.partner, runnerOptions).printCompactDiffFromZip(options.fromZip);
+      return;
+    }
 
     // If an existing sampler ID is provided, fetch and display results
     if (options.id) {
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -519,7 +519,7 @@ program
 program
   .command("run-sampler", { hidden: true})
   .description("Run Liquid Sampler for partner templates (reconciliation texts, account detail templates, and/or shared parts)")
-  .requiredOption("-p, --partner <partner-id>", "Specify the partner to be used")
+  .option("-p, --partner <partner-id>", "Specify the partner to be used (required unless --from-zip is used)")
   .option("-h, --handle <handles...>", "Specify reconciliation text handle(s) - can specify multiple")
   .option("-at, --account-template <names...>", "Specify account detail template name(s) - can specify multiple")
   .option("-s, --shared-part <names...>", "Specify shared part name(s) - can specify multiple")
@@ -554,6 +554,13 @@ program
     if (options.fromZip) {
       await new LiquidSamplerRunner(options.partner, runnerOptions).printCompactDiffFromZip(options.fromZip);
       return;
+    }
+
+    // Every other path talks to the partner/sampler API, so -p/--partner is
+    // required there - it's just not required for --from-zip's offline path.
+    if (!options.partner) {
+      consola.error("You need to specify a partner using -p or --partner");
+      process.exit(1);
     }
 
     // If an existing sampler ID is provided, fetch and display results
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -546,7 +546,7 @@ program
   .option("--compact", "Download the result and print a compact diff (named_results/results, dependencies, vanished renders, visual-only changes) grouped by template - review-friendly and safe in CI")
   .option(
     "--add-diffs-folder",
-    "With --from-zip: add a diffs/ folder to the zip containing before/after pairs only for the entries the compact diff flagged (data/scope/vanished-output/visual-only changes), instead of every sampled entry"
+    "With --from-zip: add a diffs/ folder to the zip containing before/after pairs for the entries the compact diff flagged (data/scope/vanished-output/visual-only changes) whose renders actually differ, instead of every sampled entry"
   )
   .action(async (options) => {
     // Commander sets options.open = false when --no-open is passed.
--- a/bin/cli.js
+++ b/bin/cli.js
@@ -544,15 +544,28 @@ program
   )
   .option("--no-open", "Do not download/open the report locally; only print its URL (default in CI)")
   .option("--compact", "Download the result and print a compact diff (named_results/results, dependencies, vanished renders, visual-only changes) grouped by template - review-friendly and safe in CI")
+  .option(
+    "--add-diffs-folder",
+    "With --from-zip: add a diffs/ folder to the zip containing before/after pairs only for the entries the compact diff flagged (data/scope/vanished-output/visual-only changes), instead of every sampled entry"
+  )
   .action(async (options) => {
     // Commander sets options.open = false when --no-open is passed.
     // In CI, never open regardless of the flag.
     const runnerOptions = { openReport: options.open && !process.env.CI, compact: options.compact || false };
 
+    if (options.addDiffsFolder && !options.fromZip) {
+      consola.error("--add-diffs-folder requires --from-zip <path>");
+      process.exit(1);
+    }
+
     // A local zip needs no partner API access at all - it's pure offline
     // re-analysis of a result someone already has on disk.
     if (options.fromZip) {
-      await new LiquidSamplerRunner(options.partner, runnerOptions).printCompactDiffFromZip(options.fromZip);
+      const runner = new LiquidSamplerRunner(options.partner, runnerOptions);
+      await runner.printCompactDiffFromZip(options.fromZip);
+      if (options.addDiffsFolder) {
+        runner.addDiffsFolderToZip(options.fromZip);
+      }
       return;
     }
 
--- a/lib/liquidSamplerCompact.js
+++ b/lib/liquidSamplerCompact.js
@@ -5,20 +5,28 @@ const YAML = require("yaml");
 /**
  * Extract a compact, LLM/review-friendly view of a Liquid Sampler result.
  *
- * A sampler `results.zip` is huge (the rendered HTML report alone is ~470K tokens
- * at scale, and the raw per-entry output is ~150 MB). The only signal a reviewer
- * usually needs first is the DATA diff: how each template's `named_results`
- * changed. That lives in `registers.json` for every sampled entry, before/after.
+ * A sampler `results.zip` is huge (the rendered HTML report alone can be tens of
+ * MB, and the raw per-entry output is ~150 MB at scale). This module builds
+ * several small, targeted signals instead of a raw diff:
  *
- * This module reads an already-EXTRACTED results directory (the layout inside
- * results.zip) and returns just the `named_results` changes, grouped by template
- * and deduped by identical change, so the whole thing is ~900 tokens instead of
- * hundreds of thousands. It never touches the rendered HTML.
+ *  - `templates`      - the `named_results`/`results` DATA diff, deduped by
+ *                       identical change and capped so one broken/verbose
+ *                       template can't blow up the whole summary.
+ *  - `collapsedEntries` - entries whose rendered output vanished entirely
+ *                       between phases (a strong broken-template signal that
+ *                       would otherwise show up as dozens of individual
+ *                       "value -> undefined" lines).
+ *  - `scopeTemplates` - `dependencies` / `rollforward_params` /
+ *                       `required_keys_missing` changes: what the template
+ *                       depends on, not what it renders.
+ *  - `visualOnlyEntries` - entries where `view.html` changed but the DATA
+ *                       (named_results/results) didn't - a rendering-only
+ *                       regression the data diff can't see.
  *
  * Expected directory layout (inside results.zip):
  *   <dir>/sample_entry_ids.yml
- *   <dir>/output/account_entries/<id>/{before,after}/registers.json
- *   <dir>/output/reconciliation_entries/<id>/{before,after}/registers.json
+ *   <dir>/output/account_entries/<id>/{before,after}/{registers.json,view.html}
+ *   <dir>/output/reconciliation_entries/<id>/{before,after}/{registers.json,view.html}
  */
 
 const ENTRY_KINDS = ["account_entries", "reconciliation_entries"];
@@ -27,15 +35,35 @@ const ENTRY_KINDS = ["account_entries", "reconciliation_entries"];
 // that distinctly from an explicit JSON `null`.
 const ABSENT = Symbol("absent");
 
+// Individual before/after values longer than this are elided - long free-text
+// fields (accounting-policy paragraphs, notes) are the single biggest driver of
+// oversized summaries, and the full text rarely helps a reviewer decide
+// "is this intended?" faster than a preview does.
+const MAX_VALUE_CHARS = 100;
+// Detail lines shown per template before collapsing the rest into "+N more".
+const MAX_CHANGES_SHOWN = 8;
+// Named-results/results keys that flip from a value to `undefined` in a single
+// entry, at or above this count, are treated as "the render broke" rather than
+// as that many independent findings.
+const COLLAPSE_MIN_LOST_KEYS = 3;
+// Items shown before eliding a set-diff (dependencies/rollforward params/etc.)
+const MAX_SET_ITEMS_SHOWN = 3;
+// Field-level notes shown per visual-only entry before eliding the rest.
+const MAX_VISUAL_CHANGES_SHOWN = 6;
+
 /**
  * Render a named_results value for display. Distinguishes an absent key
- * (template no longer emits it) from an explicit null.
+ * (template no longer emits it) from an explicit null, and elides long values.
  * @param {*} value
  * @returns {string}
  */
 function renderValue(value) {
   if (value === ABSENT) return "undefined";
-  return JSON.stringify(value);
+  const rendered = JSON.stringify(value);
+  if (rendered.length > MAX_VALUE_CHARS) {
+    return `${rendered.slice(0, MAX_VALUE_CHARS)}…(${rendered.length} chars)`;
+  }
+  return rendered;
 }
 
 /**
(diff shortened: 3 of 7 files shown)


lib/liquidSamplerCompact.js
```diff
@@ -131,22 +166,410 @@ function diffNamedResults(before, after) {
...
+
+/**
+ * @param {Object} deps - a `dependencies` register value
+ * @returns {Array<string>} every ledger id the template touches, from either
+ *   the flat `ledgers` list or the handles-per-ledger map
+ */
+function dependencyLedgers(deps) {
+  if (!deps) return [];
+  const handlesPerLedger = dependencyHandlesPerLedger(deps);
+  return [...new Set([...(deps.ledgers || []).map(String), ...Object.keys(handlesPerLedger)])];
```
michieldegezelle: 🟠 **Major** — A single entry with a non-array `dependencies.ledgers` (or `.account_ranges`) value throws an uncaught `TypeError` here (`.map` isn't a function on a string/number/object), which propagates out of `extractCompact` and aborts the compact diff for *every* sampled entry in the run — not just the one malformed entry. That undermines this PR's own per-entry resilience pattern already used for unreadable `registers.json` (`entriesSkipped`).

```js
function dependencyLedgers(deps) {
  if (!deps) return [];
  const handlesPerLedger = dependencyHandlesPerLedger(deps);
  const ledgers = Array.isArray(deps.ledgers) ? deps.ledgers : [];
  return [...new Set([...ledgers.map(String), ...Object.keys(handlesPerLedger)])];
}
```

lib/liquidSamplerCompact.js
```diff
@@ -158,101 +581,219 @@ function extractCompact(resultsDir) {
...
 /**
- * Format the compact diff as Markdown, suitable for stdout or a PR comment.
+ * A markdown link to an example entry: the live app URL when known, else a
+ * relative path into the results directory so a reviewer with the zip open
+ * locally still has somewhere to look.
+ * @param {string|null} url
+ * @param {string} entryKey - "kind/entryId"
+ * @returns {string}
+ */
+function exampleRef(url, entryKey) {
```
michieldegezelle, line 832: 🟡 **Minor** — `exampleRef` interpolates `url` (sourced from `sample_entry_ids.yml`) straight into a Markdown link with no scheme/format validation. With the new `--from-zip <path>`, that yml is no longer guaranteed to come from Silverfin's own sampler backend — it's whatever local zip the caller points at. A crafted `url` value can break the Markdown link syntax or inject arbitrary link text into a diff that gets posted verbatim as a PR/CI comment.

```js
function exampleRef(url, entryKey) {
  const safeUrl = typeof url === "string" && /^https?:\/\//.test(url) ? url : null;
  return safeUrl ? `([example](${safeUrl}))` : `(\`output/${entryKey}/\`)`;
}
```

lib/liquidSamplerCompact.js
```diff
@@ -131,22 +166,410 @@ function diffNamedResults(before, after) {
...
+
+  for (const m of html.matchAll(/<textarea\b([^>]*)>([\s\S]*?)<\/textarea>/g)) {
+    const name = getAttr(m[1], "data-name");
+    if (!name) continue;
+    fields.set(name, { tag: "textarea", value: decodeEntities(m[2].trim()), placeholder: getAttr(m[1], "placeholder") });
+  }
+  for (const m of html.matchAll(/<input\b([^>]*?)\/?>/g)) {
+    const name = getAttr(m[1], "data-name");
+    if (!name) continue;
+    fields.set(name, { tag: "input", value: getAttr(m[1], "value"), placeholder: getAttr(m[1], "placeholder") });
```
michieldegezelle: 🟡 **Minor** — Unlike the `textarea` (line 412) and `select` (line 423) branches just above, this `input` branch doesn't call `decodeEntities()` on `value`/`placeholder`, so an HTML-entity-encoded input value (e.g. `&amp;`) shows up un-decoded in the visual-diff output — inconsistent with the other field types.

```js
const rawValue = getAttr(m[1], "value");
fields.set(name, { tag: "input", value: rawValue !== null ? decodeEntities(rawValue) : null, placeholder: getAttr(m[1], "placeholder") });
```

lib/liquidSamplerCompact.js
```diff
@@ -131,22 +166,410 @@ function diffNamedResults(before, after) {
...
+/**
+ * Format a `results` register value for display. Flag-shaped arrays (every
+ * element 0/1) render as a triggered-count, since that's the reviewable
+ * signal (e.g. "how many unreconciled indicators fired"); anything else
+ * (raw numeric values from other template kinds) falls back to the plain
+ * rendered value.
+ * @param {*} value
+ * @returns {string}
+ */
+function formatResultsValue(value) {
```
michieldegezelle: 🟡 **Minor** — An explicit `null` `results` value and a genuinely absent/removed register both render identically as `"undefined"` here. Since `extractCompact`'s "output vanished" collapse heuristic keys off that same rendered string (`vanishedCount = dataChanges.filter(c => c.after === "undefined").length`), a template that legitimately sets `results` to `null` can get miscounted toward the "output vanished" grouping alongside genuinely lost keys — even though `named_results` already keeps this distinction via the `ABSENT` sentinel.

Illustrative fix (also needs `diffResultsRegister`'s before/after normalization updated to preserve the distinction):
```js
function formatResultsValue(value) {
  if (value === undefined) return "undefined";
  if (value === null) return "null";
  if (isFlagArray(value)) {
    const triggered = value.filter((v) => v === "1.0" || v === "1").length;
    return `${triggered}/${value.length} triggered`;
  }
  return renderValue(value);
}
```

lib/liquidSamplerCompact.js
```diff
@@ -131,22 +166,410 @@ function diffNamedResults(before, after) {
...
+    }
+  }
+}
+
+/**
+ * A stable string key identifying a change's shape, for cross-entry dedup.
+ * @param {{key: string, before?: string, after?: string, summary?: string, subLines?: Array<string>}} change
+ * @returns {string}
+ */
+function changeDedupKey(change) {
```
michieldegezelle, line 642: 💡 **Suggestion** — The dedup key's `before`/`after` components are already-truncated `renderValue()` output (capped at `MAX_VALUE_CHARS`, 100). Two distinct entries whose values share the same first 100 characters *and* the same total JSON-stringified length collapse into one deduped change with an inflated `count`, silently hiding that they're actually different values — narrow, but it cuts against this PR's own stated goal of "always disclosing what was elided rather than silently dropping it." Building the dedup key from the untruncated raw values (before rendering) would avoid this.

lib/liquidSamplerCompact.js
michieldegezelle: **Stale:** Fixed in 3aa9484 — `dependencyLedgers`/`diffDependencies` now go through a new `asArray()` helper (and `diffStringSet` hardens its own `Array.isArray` check), so a non-array `ledgers`/`account_ranges` is treated as empty instead of throwing. Covered by a new test ("does not throw when `dependencies.ledgers` or `.account_ranges` is malformed"). No action needed.

bin/cli.js
```diff
@@ -530,15 +530,32 @@ program
...
   .action(async (options) => {
     // Commander sets options.open = false when --no-open is passed.
     // In CI, never open regardless of the flag.
     const runnerOptions = { openReport: options.open && !process.env.CI, compact: options.compact || false };
 
+    // A local zip needs no partner API access at all - it's pure offline
+    // re-analysis of a result someone already has on disk.
+    if (options.fromZip) {
+      await new LiquidSamplerRunner(options.partner, runnerOptions).printCompactDiffFromZip(options.fromZip);
+      return;
```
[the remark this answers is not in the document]
michieldegezelle: **Stale:** Fixed in 3aa9484 — `-p/--partner` is now a plain `.option()`, with a runtime `if (!options.partner)` check placed after the `--from-zip` early return, so the offline path no longer needs a partner id while every other path still requires one. Covered by two new tests in `tests/bin/cli.test.js`. No action needed.

lib/liquidSamplerCompact.js
michieldegezelle: **Stale:** The scheme-validation part of this is fixed in 3aa9484 — `exampleRef` now rejects anything not matching `/^https?:\/\//`. CodeRabbit has since flagged a follow-up gap in the same fix (delimiter injection even on a scheme-valid URL) in a separate thread below — that one's still open, not covered here.

lib/liquidSamplerCompact.js
michieldegezelle: **Stale:** Fixed in 3aa9484 — the `<input>` branch now calls `decodeEntities()` on `value` (guarded for the null-attribute case), consistent with `textarea`/`select`. Covered by a new test ("decodes HTML entities in an <input> value"). No action needed.

lib/liquidSamplerCompact.js
michieldegezelle: **Stale:** Fixed in 3aa9484 — `formatResultsValue` now distinguishes `undefined` ("undefined") from an explicit `null` ("null"), and `diffResultsRegister` passes the raw before/after values through rather than a null-normalized copy. Covered by new tests. No action needed.

lib/liquidSamplerCompact.js
michieldegezelle: **Stale:** Fixed in 3aa9484 — `renderValue`'s truncation suffix now appends an 8-char sha1 hash of the full rendered string, not just its length, so two distinct long values with the same 100-char prefix and length no longer produce the same dedup key. Covered by a new test asserting distinct fingerprints. No action needed.

lib/liquidSamplerCompact.js
```diff
@@ -158,101 +618,235 @@ function extractCompact(resultsDir) {
...
+ * whatever local zip the caller points at. It's validated as an http(s) URL
+ * before being interpolated into Markdown link syntax, since this compact
+ * diff is often posted verbatim as a PR/CI comment.
+ * @param {string|null} url
+ * @param {string} entryKey - "kind/entryId"
+ * @returns {string}
+ */
+function exampleRef(url, entryKey) {
+  const safeUrl = typeof url === "string" && /^https?:\/\//.test(url) ? url : null;
+  return safeUrl ? `([example](${safeUrl}))` : `(\`output/${entryKey}/\`)`;
```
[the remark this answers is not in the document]
michieldegezelle: Fixed in cb345b6 — added a shared `sanitizeUrl()` that requires a parseable http(s) URL and rejects parentheses/whitespace/angle brackets/control characters outright (rather than escaping), and reused it for the visual-only section's "open in app" link too, which was previously interpolating `entry.url` with no validation at all. Covered by two new tests.

lib/liquidSamplerRunner.js
```diff
@@ -327,22 +327,121 @@ class LiquidSamplerRunner {
...
+        const [kind, entryId] = entryKey.split("/");
+        for (const phase of ["before", "after"]) {
+          const viewHtmlPath = path.join(resultsDir, "output", kind, entryId, phase, "view.html");
+          if (!fs.existsSync(viewHtmlPath)) continue;
+          zip.addFile(`diffs/${kind}/${entryId}/${phase}/view.html`, fs.readFileSync(viewHtmlPath));
+          filesAdded += 1;
+        }
+      }
+
+      zip.writeZip(zipPath);
```
michieldegezelle: 🟠 **Major** — `AdmZip.writeZip` rewrites `zipPath` via `open("w")` (truncate-then-write), so a mid-write failure (ENOSPC, kill, crash) permanently corrupts the original `results.zip`. For archives that are often ~150 MB and expensive to regenerate (30–60 min sampler run), that's a realistic data-loss edge case — and the intact original is still sitting in `zipBuffer` at this point.

**Fix:** write to a sibling temp file, then atomically replace:

```js
const tmpPath = `${zipPath}.tmp-${process.pid}`;
try {
  zip.writeZip(tmpPath);
  fs.renameSync(tmpPath, zipPath);
} catch (error) {
  try { fs.rmSync(tmpPath, { force: true }); } catch (_) { /* ignore */ }
  throw error;
}
```

lib/liquidSamplerRunner.js
```diff
@@ -327,22 +327,121 @@ class LiquidSamplerRunner {
...
+      consola.success(
+        `Added diffs/ folder to ${zipPath}: ${filesAdded} view.html file(s) across ${data.diffEntryKeys.length} ${
+          data.diffEntryKeys.length === 1 ? "entry" : "entries"
+        }.`
+      );
```
michieldegezelle, line 458: 🟡 **Minor** — If every flagged entry is missing `view.html` (data/scope/vanished tiers don't require it), `filesAdded` stays 0 but we still `writeZip` and print a success that claims a `diffs/` folder was added (`0 view.html file(s) across N entries`). Prefer an early info/warn return when `filesAdded === 0`, matching the empty-`diffEntryKeys` branch above — skip the rewrite entirely.

michieldegezelle (commented):
🟡 **Minor** — `CHANGELOG.md` `[1.58.0]` documents the `--compact` expansions and `--from-zip`, but the new `--add-diffs-folder` option from 07534f9 is missing. Worth a sentence on the 1.58.0 entry so the release notes match the user-facing surface (hidden command, but this repo still changelogs those).

lib/liquidSamplerRunner.js
michieldegezelle: Fixed in 5accd6b — writeZip now writes to a sibling temp file and renames into place, with cleanup on failure in the finally block.

lib/liquidSamplerRunner.js
michieldegezelle: Fixed in 0bf8f15 — added an early return (mirroring the empty-diffEntryKeys case above) when filesAdded === 0, so the zip is left untouched and no misleading success message is printed. Covered by a new test.

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 All notable changes to this project will be documented in this file.
 
+## [1.58.0] (30/07/2026)
+Expand `silverfin run-sampler --compact`: diff the `results` register, group vanished output per template instead of listing every entry, add a scope-change tier (dependencies/rollforward/required keys) and a visual-only tier for `view.html` changes the data diff can't see, and truncate/cap long output. Also adds `run-sampler --from-zip <path>` to build the compact diff from an already-downloaded `results.zip`.
```
michieldegezelle, line 6: 💡 **Suggestion** — The `1.58.0` CHANGELOG entry documents `--from-zip` but not the new `--add-diffs-folder` flag added in the same PR (`run-sampler --from-zip <path> --add-diffs-folder`). Worth a clause so release notes match the CLI surface.

BenjaminLangenakenSF (changes requested):
I found some issues with Claude, and asked for pre-made prompts that you can easily verify with your AI tool of choice 😉

---

**1. Flag-array reorders render as an identical no-op line — and dedupe into each other**

In `lib/liquidSamplerCompact.js`, `formatResultsValue` (line 203) reduces a flag-shaped `results` vector to a triggered *count*. So a genuine change where the same number of indicators fire but different ones do — `["1.0","0.0"] → ["0.0","1.0"]` — renders as:

```
- `results`: `1/2 triggered` → `1/2 triggered`
```

`diffResultsRegister` correctly detects it as a change (it compares the raw JSON), but the rendered line tells a reviewer nothing, and reads like a bug in the diff tool itself.

It's worse than cosmetic: that rendered string doubles as the cross-entry dedup key in `changeDedupKey` (line 565). Two entries of the same template with *opposite* reorders collapse into one `[2×] results: 1/2 triggered → 1/2 triggered` line — two different real changes reported as one repeated change. Verified both by running `extractCompact`/`formatCompact` over a two-entry fixture with mirrored flag vectors.

This is exactly the failure mode the `#<hash>` fingerprint in `renderValue` was added to prevent for truncated long values (see that function's own comment about dedup keys) — the same guard is just missing on the flag-array path. Either name the flipped indices when the counts match but the vectors differ, or append a short hash of the vector the way `renderValue` does.

---

**2. The `--add-diffs-folder` success message overstates what landed in the zip**

In `lib/liquidSamplerRunner.js:426`, the success line is built from `data.diffEntryKeys.length` — every entry the compact diff flagged — not from the entries that actually got a `view.html` written. `filesAdded` is tracked correctly right above it (lines 405-412) but only used for the count of *files*.

On a run where 9 entries were flagged but only 2 of them had a `view.html` on disk, it printed:

```
✔ Added diffs/ folder to results.zip: 4 view.html file(s) across 9 entries.
```

`diffs/` contained 2 entries. That's the normal case, not an edge case — the data and scope tiers flag plenty of entries whose `view.html` was never rendered, and a reviewer who opens `diffs/` expecting 9 finds 2 and assumes the folder is broken. Count the distinct entries that contributed a file instead (track a `Set` alongside `filesAdded`), and keep the `entry`/`entries` pluralisation keyed off that number.

---

**3. `--add-diffs-folder` is missing from the CHANGELOG**

The 1.58.0 entry in `CHANGELOG.md` documents `run-sampler --from-zip` but says nothing about `--add-diffs-folder`, which arrived in the three commits after that entry was written (`a78b80a`, `2f2f2f4`, `0bf8f15`). The CHANGELOG is surfaced to users by the CLI itself (`lib/cli/changelogReader.js`), so a new user-facing flag that isn't in it is effectively undocumented. Add a sentence covering it — worth mentioning that it only adds before/after pairs for the entries the compact diff flagged, since that's the whole point versus extracting the full archive.

---

**4. Two small ones**

Two non-blocking notes on the new code:

1. `extractNamedFields` in `lib/liquidSamplerCompact.js` keys fields by `data-name` into a `Map`, so a repeated `data-name` (a radio group: several `<input>`s sharing one name) is last-wins, and the `checked` attribute isn't read at all — only `value`. A change in which radio is selected therefore produces no field-level note and falls through to the generic *"layout/markup changed with no anchored field explaining it"* line. That fallback is honest, so this isn't a correctness bug, but if radio groups are common in the sampled templates it's a blind spot worth either fixing (read `checked`, key by `name + value`) or naming in the function's doc comment alongside the existing note about fields without a `data-name`.

2. `--from-zip --add-diffs-folder` extracts the zip and runs `extractCompact` over it twice — once in `printCompactDiffFromZip` (line 357) and again in `addDiffsFolderToZip` (line 384), each with its own `mkdtemp`/`rmSync` cycle. Correct, just wasteful; `diffEntryKeys` from the first pass could be handed to the second.

BenjaminLangenakenSF (changes requested):
Small pushback:

1.. The mirrored-reorder collapse survives. flipped is the set of differing indices, which is the same computed from either direction, so both sides render 1/2 triggered (indices 0,1 flipped) — identical strings, and changeDedupKey derives from them. Two entries with opposite reorders still dedupe into one [2×] line. The new test asserts the different-index-set case, which is why it passes. Naming which indices went 0→1 vs 1→0 fixes the rendered line and the key together.

Also: the suffix is uncapped, and the flag-array branch never reaches renderValue, so neither MAX_VALUE_CHARS nor the #<hash> guard applies. A long vector with many flips renders a 200+ char line.

Human readable 😅:
The fix that works: say which way each switch went — "0 turned off, 1 turned on" — so the two directions print differently and stop being mistaken for each other.

4.1. /\bchecked\b/ matches aria-checked and data-checked — \b matches after a hyphen. Once every radio in a group looks checked, last-wins returns and the selection change is invisible again. Anchor on preceding whitespace. Same weakness in getAttr(m[1], "type") vs data-type="radio". I couldn't tell from the repo whether the sampled markup emits aria-checked, so this may be latent rather than live.

Human readable 😅:
If every button in a group carries aria-checked, they all look selected, the code falls back to just keeping the last one, and a changed selection goes unnoticed — back to the original bug the fix was meant to solve.
The fix is to only accept checked when there's a space in front of it, not a hyphen.

Everything else ok 😉

lib/liquidSamplerCompact.js
```diff
@@ -131,22 +178,514 @@ function diffNamedResults(before, after) {
...
+ * @param {*} before
+ * @param {*} after
+ * @returns {{before: string, after: string}|null}
+ */
+function diffResultsRegister(before, after) {
+  const b = before == null ? null : before;
+  const a = after == null ? null : after;
+  if (JSON.stringify(b) === JSON.stringify(a)) return null;
+  const suffix = describeFlagFlipSuffix(before, after);
+  return { before: formatResultsValue(before, suffix), after: formatResultsValue(after, suffix) };
```
BenjaminLangenakenSF: The suffix goes to both sides, so `before` and `after` still render as the same string within a line - it reads like a no-op even though the direction is now in the text. Dropping it from the `before` side fixes that and keeps the mirrored dedup keys distinct: checked against the patched module, one entry renders `1/2 triggered` -> `1/2 triggered (0 off, 1 on)` and its mirror `(0 on, 1 off)`.

It also makes the doc comment at lines 210-212 true as written - it currently claims a same-count reorder doesn't render as an identical string on both sides, which is exactly what it does.

```suggestion
  return { before: formatResultsValue(before), after: formatResultsValue(after, suffix) };
```

BenjaminLangenakenSF (approved):
Approved!

Two non-blocking nitpicks:
- The first has a suggestion you can apply directly
- Test file, expect(diffA.before.length).toBeLessThan(100): Only bites if you apply the suggestion above: once before drops the suffix this just measures "10/20 triggered" and stops testing the cap. diffA.after keeps it meaningful.

